PR #7, opened 2020-06-03
Add dev_utils

jbygdell:
* certificates
* c4gh keys
* dummy data
  * The first line is "THIS FILE IS JUST DUMMY DATA" followed by a 1Mib binary blob. 
* configuration files
* docker-compose file
  *  to start the backend services (MQ, DB, S3)


dtitov:
Why do we have RabbitMQ configuration in this repo? Can't we just refer the existing broker image?

jbygdell:
> Why do we have RabbitMQ configuration in this repo? Can't we just refer existing broker image?

Not if you want to fake the CEGA side without the shoveling stuff

PR #10, opened 2020-06-04
Basic file verification

jbygdell:
Any optimization should be done separately at a later date.

* `GetHeader` retrieves the header from the DB and returns it as a byte array
* `MarkCompleted` marks the file as 'COMPLETED' in the DB
* `FileReader` returns the file as `io.Read`, it needs to be extended to return the data stream when reading a file from S3

The read, decrypt, sha part should later on be placed in a for loop so that we only read parts of the file at the time in order to limit memory consumption.


cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
-	var message Message
-	// var header []byte
 
 	go func() {
 		for d := range broker.GetMessages(mq, config.Broker.Queue) {
 			log.Debugf("received a message: %s", d.Body)
-			err := json.Unmarshal(bytes.Replace(d.Body, []byte(`'`), []byte(`"`), -1), &message)
-			if err != nil {
+			var m Message
+			e := json.Unmarshal(bytes.Replace(d.Body, []byte(`'`), []byte(`"`), -1), &m)
```
dtitov, line 63: Why are you replacing single quotes with double quotes? If it's a valid JSON, it should be already with double-quotes. If it has single quotes, it should be treated as a malformed JSON, AFAIK.

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+			if e != nil {
 				log.Errorf("Not a json message: %s", err)
+				// Nack errorus message so the server gets notified that something is wrong but don't requeue the message
+				d.Nack(false, false)
```
dtitov, line 67: Here and below you are ignoring error from Nack() method. shouldn't it be processed somehow (at least, logged)? If you don't want to do anything about it, ignore it explicitly with `_`, following the Go style.

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
 	log.Info("starting verify service")
-	var message Message
-	// var header []byte
 
 	go func() {
 		for d := range broker.GetMessages(mq, config.Broker.Queue) {
 			log.Debugf("received a message: %s", d.Body)
-			err := json.Unmarshal(bytes.Replace(d.Body, []byte(`'`), []byte(`"`), -1), &message)
-			if err != nil {
+			var m Message
```
dtitov, line 62: Can we give slightly more meaningful names to variables (here and in other places)? I mean, `d`, `m`?

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+				// Nack errorus message so the server gets notified that something is wrong but don't requeue the message
+				d.Nack(false, false)
+				// Send the errorus message to an error queue so it can be analyzed.
+				if err := broker.SendMessage(mq, d.CorrelationId, config.Broker.Exchange, config.Broker.RoutingError, d.Body); err != nil {
+					log.Error("faild to publish message, reason: ", err)
+				}
+				// Restart on new message
+				continue
+			} else {
+				e := reflect.ValueOf(&m).Elem()
```
dtitov, line 75: I'm not sure what is this reflection magic for... Are you validating unmarshalled JSON like that? If struct-fields are not pointers and they are missing in the JSON, unmarshalling will fail anyway. Also, it's better to validate the JSON message against the JSON schema before unmarshalling, IMHO (if you want to have validation at all).

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+					if e.Field(i).Interface() == "" || e.Field(i).Interface() == 0 {
+						log.Errorf("%s is missing", e.Type().Field(i).Name)
+						err = fmt.Errorf("%s is missing", e.Type().Field(i).Name)
+					}
+				}
+				if err != nil {
+					// Nack errorus message so the server gets notified that something is wrong but don't requeue the message
+					d.Nack(false, false)
+					// Send the errorus message to an error queue so it can be analyzed.
+					if err := broker.SendMessage(mq, d.CorrelationId, config.Broker.Exchange, config.Broker.RoutingError, d.Body); err != nil {
```
dtitov, line 86: Here and below: what's the actual error message published to the queue? It looks like you are just publishing original message there. Shouldn't some additional details be added to it (like it's done now in Python)?

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+				log.Error(e)
+			}
+			defer k.Close()
+
+			key, e := keys.ReadPrivateKey(k, []byte(config.Crypt4gh.Passphrase))
+			if e != nil {
+				log.Error(e)
+			}
+
+			f, err := os.Open(filepath.Join(filepath.Clean(config.Archive.Location), m.ArchivePath))
```
dtitov, line 118: `k`, `e`, `f`?

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+			}
+			hash := sha256.New()
+			if _, err := io.Copy(hash, c4ghr); err != nil {
+				log.Error(err)
+			}
+			key = [32]byte{}
+
+			// Mark file as "COMPLETED"
+			if e := postgres.MarkCompleted(db, fmt.Sprintf("%x", hash.Sum(nil)), m.FileID); e != nil {
+
```
dtitov, line 144: So what do we do if `err != nil`?

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+				log.Error(err)
+			}
+			hash := sha256.New()
+			if _, err := io.Copy(hash, c4ghr); err != nil {
+				log.Error(err)
+			}
+			key = [32]byte{}
+
+			// Mark file as "COMPLETED"
+			if e := postgres.MarkCompleted(db, fmt.Sprintf("%x", hash.Sum(nil)), m.FileID); e != nil {
```
dtitov, line 143: So, as we discussed, verification will be also triggered separately, by some cron job. Do we need to touch the database every time upon this recurring verification?

cmd/verify/verify.go
```diff
@@ -40,40 +55,110 @@ func main() {
...
+				completed, err := json.Marshal(&c)
 				if err != nil {
-					log.Fatal(err)
+					// do something
+				}
+				if err := broker.SendMessage(mq, d.CorrelationId, config.Broker.Exchange, config.Broker.RoutingKey, completed); err != nil {
+					log.Error("faild to publish message, reason: ", err)
 				}
-				fmt.Println(header)
+				d.Ack(false)
```
dtitov, line 160: Error is unhandled.

dev_utils/config.yaml
```diff
@@ -23,6 +23,7 @@ broker:
   queue: "stableIDs"
   exchange: "localega"
   routingKey: "errmsg"
```
dtitov, line 25: Typo? Should `routingKey` be `errmsg`?

internal/storage/storage.go
```diff
@@ -1,20 +1,41 @@
...
 	GID      int
 }
+
+// FileReader returns
+func FileReader(archive, filePath string) io.Reader {
+	var r io.Reader
+	if archive == "s3" {
+		// s3 specifc stuff
+	} else {
+		f, err := os.Open(filePath)
```
dtitov, line 34: `r`, `f`

internal/postgres/postgres.go
```diff
@@ -64,3 +65,35 @@ func buildConnInfo(c Pgconf) string {
...
+}
+
+// MarkCompleted markes the file as "COMPLETED"
+func MarkCompleted(db *sql.DB, checksum string, fileID int) (error) {
+	const completed = "UPDATE local_ega.files SET status = 'COMPLETED', archive_file_checksum = $1, archive_file_checksum_type = 'SHA256'  WHERE id = $2;"
+	res, err := db.Exec(completed, checksum, fileID)
+	if err != nil {
+		log.Errorf("something went wrong with the DB qurey: %s", err)
+	}
+	if ra, _ := res.RowsAffected(); ra == 0 {
```
dtitov, line 95: Sorry for nitpicking, but, again, `ra`, res`? Can we have it more detailed? It's clear that you know the meaning while you are writing the code, but when others are reading it, we actually have to jump to the declaration of the variable to see that `ra` is actually `RowsAffected`. Why not just call it `rowsAffected`?

cmd/verify/verify.go
jbygdell: That is to ensure that the incoming message doesn't contains null values for the keys we require.
 {"user": ""} or {"user": null} are both valid json and the unmarshaling will not catch that.

Once we have the proper message definitions we could use schema validation but for now this is it.

dtitov: If it's `""`, then it's a valid string - just empty. And we can check if it's empty without reflection. But are you sure that `null`s will not fail during unmarshalling? I think it will if this string is not a pointer in the struct. But I will double-check.

cmd/verify/verify.go
jbygdell: Strange that that one didn't raise a flag, normally the code tools will flag that sort of stuff.

dtitov: Which IDE do you use? I use Goland and it highlights it :)

dev_utils/config.yaml
jbygdell: It can be whatever you want.

cmd/verify/verify.go
jbygdell: Then we need something in the message struct that tells the verifier NOT to touch the database

dtitov: Yes. But you can already add this flag to the message struct. If it's not there, it will be `null` (make it pointer). If it's there, then don't write to DB.

cmd/verify/verify.go
jbygdell: That was  necessary in February but apparently not now, might have to do with the update to GO 1.14

cmd/verify/verify.go
jbygdell: Running the unmarshalling on the second example above does not generate an error.

PR #11, opened 2020-06-04
Use struct to wrap db instance

jonandernovella:
Use db struct instead of the db object directly.


cmd/finalize/finalize.go
```diff
@@ -27,9 +28,10 @@ type Message struct {
 func main() {
 	config := NewConfig()
 	mq := broker.New(config.Broker)
-	db, err := postgres.NewDB(config.Postgres)
-	if err != nil {
-		log.Println("err:", err)
+	dbs := postgres.NewDB(config.Postgres)
+	db := dbs.Db
+	if dbs.Error != nil {
```
viklund, line 33: You don't need to check the error here. The NewDB function checks all errors and panics if something stupid happens.

viklund:
> Actually what you only need to do is to remove the `SQLdb` struct

From my point of view, this is a small first step to have a db interface object that itself does all the SQL. So the SQLdb struct should take methods like `.SetStableId` and so on.

internal/postgres/postgres.go
```diff
@@ -12,7 +12,8 @@ import (
 
 // SQLdb is
 type SQLdb struct {
-	Db *sql.DB
+	Db    *sql.DB
+	Error error
```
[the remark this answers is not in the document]
dtitov: Agree, errors should not be swallowed like this.

cmd/finalize/finalize.go
dtitov: Not panics, but rather returns `nil` DB along with the `error`. At least, I personally prefer to avoid panicking and push the errors upwards instead.

dtitov:
> Actually what you only need to do is to remove the `SQLdb` struct

I would like it too, but then we need to remove MQBroker as well. Because now it serves the same purpose of incapsulating some logic within. It would be not fair to remove DB facade, but keep MQ facade. So we either remove both or keep both.

jbygdell:
> > Actually what you only need to do is to remove the `SQLdb` struct
> 
> I would like it too, but then we need to remove `MQBroker` as well. Because now it serves the same purpose of incapsulating some logic within. It wouldn't make sense to remove DB facade, but keep MQ facade. So we either remove both or keep both.

Not necessarily since they are 2 different things.
The `SQLdb` is not used right now and the reason is that we already have a set of encapsulation by the `database/sql` library, `lib/pq` is the low level library that you should compare against the `steadway/amqp` library.
The `NewDB` function already returns a complete DB struct object.

PR #16, opened 2020-06-06
Shared storage components

### LINKED ISSUE #3: shared storage components
- [x] posix support
- [ ] S3 support

jonandernovella:
This PR aims at building a modularised storage backend interface, focusing on the addition of S3 file reading and uploading support.

Fixes #3

### DIFF
--- a/cmd/verify/config.go
+++ b/cmd/verify/config.go
@@ -1,8 +1,13 @@
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"path"
+	"reflect"
 	"strings"
 
 	log "github.com/sirupsen/logrus"
@@ -25,9 +30,11 @@ var (
 
 // Config is a parent object for all the different configuration parts
 type Config struct {
-	Archive  storage.Conf
-	Broker   broker.Mqconf
-	Crypt4gh struct {
+	ArchiveType  string
+	PosixArchive storage.PosixConf
+	S3Archive    storage.S3Conf
+	Broker       broker.Mqconf
+	Crypt4gh     struct {
 		KeyPath    string
 		Passphrase string
 	}
@@ -147,8 +154,7 @@ func (c *Config) readConfig() {
 	}
 
 	if viper.GetString("archive.type") == "s3" {
-		s3 := storage.Conf{}
-		s3.Type = "s3"
+		s3 := storage.S3Conf{}
 		// All these are required
 		s3.URL = viper.GetString("archive.url")
 		s3.AccessKey = viper.GetString("archive.accesskey")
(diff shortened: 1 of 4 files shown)


cmd/verify/verify.go
```diff
@@ -49,6 +49,14 @@ func main() {
 		log.Println("err:", err)
 	}
 
+	// Storage logici for S3 or Posix
+	var backend storage.Backend
+	if config.ArchiveType == "posix" {
+		backend = storage.NewPosixBackend(config.PosixArchive)
+	} else {
+		backend = storage.NewS3Backend(config.S3Archive, TransportConfigS3(config))
+	}
```
jbygdell, line 58: This is not the way to do this.
With the logic in the proper place you will be able to do it this way:

`archive := storage.NewBacked(config.Storage)`

Which would mean that the call later on in this file would look like this:

`f := archive.ReadFile(filepath.Join(filepath.Clean(config.PosixArchive.Location), message.ArchivePath)`

cmd/verify/verify.go
```diff
@@ -102,7 +110,13 @@ func main() {
...
 				log.Error(err)
 			}
-			f := storage.FileReader(config.Archive.Type, filepath.Join(filepath.Clean(config.Archive.Location), message.ArchivePath))
+
+			var f io.Reader
+			if config.ArchiveType == "posix" {
+				f = backend.ReadFile(filepath.Join(filepath.Clean(config.PosixArchive.Location), message.ArchivePath))
+			} else {
+				f = backend.ReadFile(message.ArchivePath)
+			}
```
jbygdell, line 119: See comment above.

cmd/verify/config.go
```diff
@@ -214,3 +225,35 @@ func parseConfig() {
...
+			log.Debug("no certs appended, using system certs only")
+		}
+	}
+
+	trConfig := &http.Transport{
+		TLSClientConfig:   cfg,
+		ForceAttemptHTTP2: true}
+
+	return trConfig
+}
```
jbygdell, line 259: This could be placed in the storage.go file.

go.mod
```diff
@@ -3,9 +3,13 @@ module sda-pipeline
 go 1.14
 
 require (
+	github.com/aws/aws-sdk-go v1.31.13
 	github.com/elixir-oslo/crypt4gh v0.0.2
+	github.com/go-ini/ini v1.57.0 // indirect
 	github.com/google/uuid v1.1.1
 	github.com/lib/pq v1.6.0
+	github.com/minio/minio-go v6.0.14+incompatible
+	github.com/minio/minio-go/v6 v6.0.56
```
jbygdell, line 12: Why is minio included here if it is not used?

cmd/verify/config.go
viklund: I don't really care where this is placed. But we should probably have a generic function to init tlsConfigs for everything, but that's a separate PR.

cmd/verify/verify.go
viklund: The backend here should know everything about how to construct the filepath. `verify.go` should only setup the correct backend at the start and then never care about it again, if the POSIX backend need to do some filepath joining or something, that should be done inside the `ReadFile` function so these lines can be only:

```go
f := backend.ReadFile(message.ArchivePath)
```

go.mod
jonandernovella: Fixed

cmd/verify/verify.go
viklund: We will have this config switch on backend somewhere. I think it's good to have that as early in the runtime process as possible.

Is this what you mean all you mean? To have a function in `storage.go` like this (pseudo-go because I'm not good at go without a compiler):

```go
func NewBackend(config) Backend {
 	if config.ArchiveType == "posix" {
		return storage.NewPosixBackend(config.PosixArchive)
	} else {
		return storage.NewS3Backend(config.S3Archive, TransportConfigS3(config))
	} 
}
```

cmd/verify/config.go
jonandernovella: Ok, let's fix this later!

cmd/verify/verify.go
jonandernovella: Done

go.mod
```diff
@@ -3,7 +3,9 @@ module sda-pipeline
 go 1.14
 
 require (
+	github.com/aws/aws-sdk-go v1.31.13
 	github.com/elixir-oslo/crypt4gh v0.0.2
+	github.com/go-ini/ini v1.57.0 // indirect
```
jbygdell, line 8: This doesn't seem to be needed

jonandernovella: Done

PR #17, opened 2020-06-08
Add meta linter golangci-lint

pontus:
Add workflow using https://github.com/marketplace/actions/run-golangci-lint to provide multilinting.

Runs with `only-new-issues` to not annotate old code on PR (newish feature that may not behave correctly).

### DIFF
--- a/.github/workflows/report.yaml
+++ b/.github/workflows/report.yaml
@@ -1,6 +1,12 @@
 name: Multilinters
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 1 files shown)


dtitov:
Nice plugin 👍  I would merge it, but the linter already fails for our current code. Should we merge it regardless?

pontus:
It appears that it should be configured exactly as they suggest to handle that nicer. Apologies for requesting review prematurely.

PR #19, opened 2020-06-10
Ingest

### LINKED ISSUE #2: [Ingest] copy file from inbox to archive
- [x] split the c4gh header
- [x] decrypt and validate the header

jbygdell:
Basic ingest functionality on posix backend

Fixes #2

### DIFF
--- a/cmd/ingest/ingest.go
+++ b/cmd/ingest/ingest.go
@@ -43,6 +43,19 @@ func main() {
 	config := NewConfig()
 	mq := broker.New(config.Broker)
 	db, err := postgres.NewDB(config.Postgres)
+	var archive, inbox storage.Backend
+	if config.ArchiveType == "s3" {
+		archive = storage.NewS3Backend(config.ArchiveS3, TransportConfigS3(config.ArchiveS3))
+	} else {
+		archive = storage.NewPosixBackend(config.ArchivePosix)
+	}
+
+	if config.InboxType == "s3" {
+		inbox = storage.NewS3Backend(config.InboxS3, TransportConfigS3(config.InboxS3))
+
+	} else {
+		inbox = storage.NewPosixBackend(config.InboxPosix)
+	}
 	if err != nil {
 		log.Println("err:", err)
 	}
@@ -63,29 +76,29 @@ func main() {
 				log.Errorf("Not a json message: %s", err)
 			}
 
-			fileID, err := db.InsertFile(fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath), message.User)
+			fileID, err := db.InsertFile(message.FilePath, message.User)
 			if err != nil {
 				log.Errorf("InsertFile failed, reason: %v", err)
 				// This should really be hadled by the DB retry mechanism
 			}
 
-			file, err := storage.ReadFile(config.Inbox.Type, fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath))
+			file, err := inbox.ReadFile(message.FilePath)
 			if err != nil {
-				log.Errorf("Failed to open file: %s, reason: %v", fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath), err)
+				log.Errorf("Failed to open file: %s, reason: %v", message.FilePath, err)
 				continue
 			}
 
-			fileSize, err := storage.GetFileSize(config.Inbox.Type, fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath))
+			fileSize, err := inbox.GetFileSize(message.FilePath)
 			if err != nil {
-				log.Errorf("Failed to get file size: %s, reason: %v", fmt.Sprintf("%s/%s", config.Archive.Location, message.FilePath), err)
+				log.Errorf("Failed to get file size of: %s, reason: %v", message.FilePath, err)
 				continue
 			}
 
 			// 4MiB readbuffer, this must be large enough that we get the entire header and the first 64KiB datablock
 			// Should be made configurable once we have S3 support
 			var bufSize int
-			if bufSize = 4 * 1024 * 1024; config.Inbox.Chunksize > 4*1024*1024 {
-				bufSize = config.Inbox.Chunksize
+			if bufSize = 4 * 1024 * 1024; config.InboxS3.Chunksize > 4*1024*1024 {
+				bufSize = config.InboxS3.Chunksize
 			}
 			readBuffer := make([]byte, bufSize)
 			hash := sha256.New()
@@ -94,9 +107,9 @@ func main() {
 
 			// Create a random uuid as file name
 			archivedFile := uuid.New().String()
-			dest, err := storage.WriteFile(config.Archive.Type, fmt.Sprintf("%s/%s", config.Archive.Location, archivedFile))
+			dest, err := archive.WriteFile(archivedFile)
 			if err != nil {
-				log.Errorf("Failed to create file: %s, reason: %v", fmt.Sprintf("%s/%s", config.Archive.Location, archivedFile), err)
+				log.Errorf("Failed to create file: %s, reason: %v", archivedFile, err)
 				continue
 			}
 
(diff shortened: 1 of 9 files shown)


cmd/ingest/ingest.go
```diff
@@ -42,24 +57,162 @@ func main() {
...
+				// This should really be hadled by the DB retry mechanism
+			}
+
+			file, rfe := storage.ReadFile(config.Inbox.Type, fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath))
+			if rfe != nil {
+				log.Errorf("Failed to open file: %s, reason: %v", fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath), rfe)
+				continue
+			}
+
+			fileSize, gfse := storage.GetFileSize(config.Inbox.Type, fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath))
```
dtitov, line 78: Why are you creating new variables for `error` every time? You can just keep reusing `err` which you had in the beginning. No need in `e`, `ife`, `rfe`, `gfse`, etc.

internal/broker/broker.go
```diff
@@ -97,7 +98,15 @@ func GetMessages(b *AMQPBroker, queue string) <-chan amqp.Delivery {
 }
 
 // SendMessage sends message to RabbitMQ if the upload is finished
-func SendMessage(b *AMQPBroker, corrID, exchange, routingKey string , body []byte) error {
+func SendMessage(b *AMQPBroker, corrID, exchange, routingKey string, body []byte) error {
+	// Set channel
+	if e := b.Channel.Confirm(false); e != nil {
```
dtitov, line 103: What is this? What is it needed for?

internal/broker/broker.go
```diff
@@ -183,3 +192,15 @@ func TLSConfigBroker(b Mqconf) *tls.Config {
 	}
 	return cfg
 }
+
+// // One would typically keep a channel of publishings, a sequence number, and a
+// // set of unacknowledged sequence numbers and loop until the publishing channel
+// // is closed.
+func confirmOne(confirms <-chan amqp.Confirmation) error {
```
dtitov, line 199: What is this? Why do we have a queue of confirmations?

cmd/ingest/ingest.go
```diff
@@ -42,24 +57,162 @@ func main() {
...
+			if gfse != nil {
+				log.Errorf("Failed to get file size: %s, reason: %v", fmt.Sprintf("%s/%s", config.Archive.Location, message.FilePath), gfse)
+				continue
+			}
+
+			hash := sha256.New()
+			// 512KiB readbuffer, this must be large enough that we get the entire header and the first 64KiB datablock 
+			readBuffer := make([]byte, 512 * 1024)
```
dtitov, line 86: After a day of consideration, I'm still a bit skeptical about the fixed-sized buffer approach. Files can be very large. For example, our test submission is 40+GB, which means that encrypted it would contain ~ 625000 data blocks. And according to Crypt4GH specification, each data block can be theoretically encrypted with a separate header packet, which would lead to ~ 625000 header packets in this case. I know, it would be a very special case, but I think that we need to respect Crypt4GH standard anyway. And this header will not fit into the remaining 512 - 64 = 448 KBs (448000 / 625000 = less than a byte per header). So, IMHO, this approach is both not very neat in terms of code, but also simply not 100% reliable. I'd argue for opening the stream twice (as was suggested in Slack).

cmd/ingest/ingest.go
jbygdell: except that you get shadow declaration warnings for every time you repeat the declaration

cmd/ingest/ingest.go
dtitov: But it can work for now, as a minimally viable solution. And can be reworked later.

cmd/ingest/ingest.go
```diff
@@ -42,24 +57,162 @@ func main() {
...
+			brokerMsg, err := json.Marshal(&msg)
+			if err != nil {
+				log.Error(err)
+				// This should really not fail.
+			}
+			if err := broker.SendMessage(mq, delivered.CorrelationId, config.Broker.Exchange, config.Broker.RoutingKey, brokerMsg); err != nil {
+				// TODO fix resend mechainsm
+				log.Errorln("We need to fix this resend stuff ...")
+			}
+			delivered.Ack(false)
```
dtitov, line 177: Ack() `error` is ignored.

cmd/ingest/ingest.go
jbygdell: In the end it will be configurable in order to match the S3 chunk size and then we are talking ~8-32MiB buffer size.

cmd/ingest/ingest.go
dtitov: This is weird. I have just replaced all the errors with `err` in this method and IDE didn't highlight anything. Neither did linter complain. Can you give a particular example of the problematic usage of `err` in this method?

jbygdell: If you have `err` on L66 you will get a shado declaration error on L111

`
go vet -vettool=$(which shadow) cmd/ingest/*
cmd/ingest/ingest.go:111:14: declaration of "err" shadows declaration at line 66
`

dtitov: First of all, in all the places where you have `_` before the `err`  on the left side you should use `=` instead of `:=` - this will reuse existing `err` and will not produce the shadowing warning. Secondly, this shadowing feature is experimental and disabled by default, reportedly having many false-positives. I don't know why would you enable it. Check this out, for example: https://github.com/golang/go/issues/19490  In third, shadowing is not necessarily bad even if it really happens. In the example above, having outer `err` from line 66 shadowed by the `inner` error at line 111 doesn't break anything. So, again, I'd rather disable this linter feature (since it's disabled by default).

dtitov: P.S. Also I don't know which IDE do you use, but in case if shadowing can really break something or have negative consequences, GoLand highlights it in the code. But if it's harmless shadowing, it doesn't.

internal/broker/broker.go
jbygdell: When writing to a durable exchange we want confirmation that the message has been persisted.

internal/broker/broker.go
jbygdell: `Confirm puts this channel into confirm mode so that the client can ensure all publishings have successfully been received by the server.`

jonandernovella: Does this block?

cmd/ingest/ingest.go
```diff
@@ -42,24 +57,170 @@ func main() {
...
+
+			fileSize, err := storage.GetFileSize(config.Inbox.Type, fmt.Sprintf("%s/%s", config.Inbox.Location, message.FilePath))
+			if err != nil {
+				log.Errorf("Failed to get file size: %s, reason: %v", fmt.Sprintf("%s/%s", config.Archive.Location, message.FilePath), err)
+				continue
+			}
+
+			// 4MiB readbuffer, this must be large enough that we get the entire header and the first 64KiB datablock 
+			// Should be made configurable once we have S3 support
+			readBuffer := make([]byte, 4 * 1024 * 1024)
```
jonandernovella, line 86: This should be fixed

internal/broker/broker.go
jbygdell: > Does this block?

block what? be more specific.

cmd/ingest/ingest.go
jbygdell: What should be fixed?

jonandernovella: The size of the bytes array should be configurable (from the comment above)

internal/broker/broker.go
dtitov: I think Jon means if it's blocking call or async one. And my guess would be that it's blocking, but I don't think that it matters.

jonandernovella: If this is written to a channel, my guess is async. But I am not sure... The python code made blocking amqp calls.

dtitov: It's blocking in terms of programming, not in terms of messaging. I mean, it blocks for a time needed to put a message to a broker, but no longer.

PR #25, opened 2020-06-11
Shared config module

### LINKED ISSUE #20: shared config code

jbygdell:
Modularized configuration that takes the application type as parameter `config.New("ingest")`
* moved S3 tls config to the storage module
* uppdated all apps to use the new config

Fixes #20


internal/config/config.go
```diff
@@ -45,10 +40,31 @@ type Crypt4gh struct {
...
+		viper.AddConfigPath(path.Join(ss...))
+	}
+
+	if viper.IsSet("configFile") {
+		viper.SetConfigFile(viper.GetString("configFile"))
+	}
+
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			log.Infoln("No config file found, using ENVs only")
```
blankdots, line 63: so it is either file or env only, no more combinations. that is ok. just wanted to double check.

jbygdell: It's and/or, env overrules file values.

PR #28, opened 2020-06-12
Fix s3 download

jbygdell:
* fixes s3 client config
* adds a `Downloader` to the S3 struct
* updates the config template so we only need to specify s3 or posix for the storage parts
* adds a s3cmd conf file for testing S3 against the minio instance in the docker-compose file.
* minor cleanup to make stuff work

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -177,7 +177,7 @@ func (c *Config) configInbox() {
 		// All these are required
 		s3.URL = viper.GetString("inbox.url")
 		s3.AccessKey = viper.GetString("inbox.accesskey")
-		s3.SecretKey = viper.GetString("arcinboxhive.secretkey")
+		s3.SecretKey = viper.GetString("inbox.secretkey")
 		s3.Bucket = viper.GetString("inbox.bucket")
 
 		if viper.IsSet("inbox.port") {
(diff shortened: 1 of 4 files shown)


internal/config/config.go
```diff
@@ -149,6 +149,12 @@ func (c *Config) configArchive() {
 			s3.Port = 443
 		}
 
+		if viper.IsSet("archive.region") {
+			s3.Region = viper.GetString("archive.region")
+		} else {
+			s3.Region = "us-east-1"
+		}
```
pontus, line 156: I'd really prefer having defaults initializations before (collected), second best would be the go pattern

`if s3.Region = "us-east-1"; viper.IsSet("archive.region") {
￼			s3.Region = viper.GetString("archive.region")
￼		}

internal/config/config.go
```diff
@@ -149,6 +149,12 @@ func (c *Config) configArchive() {
...
+
 		if viper.IsSet("archive.chunksize") {
 			s3.Chunksize = viper.GetInt("archive.chunksize") * 1024 * 1024
 		}
```
pontus, line 160: Can we have the default for `chunksize`/`bufsize` here rather than cases for it it the ingest?

internal/config/config.go
jbygdell: > Can we also join the main parts of configInbox and configArchive in and just pass the prefix to pick settings from (in internal/config/config.go) to a common function?

Changes to the config module should be done a a separate PR, but yes I see no reason not to.

PR #33, opened 2020-06-15
Fix S3 upload function

jbygdell:
For the S3 uploader to work we need to wrap it in a go routine and we must explicitly close the writer one we are done.

### DIFF
--- a/cmd/ingest/ingest.go
+++ b/cmd/ingest/ingest.go
@@ -84,7 +84,7 @@ func main() {
 				// This should really be hadled by the DB retry mechanism
 			}
 
-			file, err := inbox.ReadFile(message.FilePath)
+			file, err := inbox.NewFileReader(message.FilePath)
 			if err != nil {
 				log.Errorf("Failed to open file: %s, reason: %v", message.FilePath, err)
 				continue
--- a/cmd/ingest/ingest.go
+++ b/cmd/ingest/ingest.go
@@ -109,7 +109,7 @@ func main() {
 
 			// Create a random uuid as file name
 			archivedFile := uuid.New().String()
-			dest, err := archive.WriteFile(archivedFile)
+			dest, err := archive.NewFileWriter(archivedFile)
 			if err != nil {
 				log.Errorf("Failed to create file: %s, reason: %v", archivedFile, err)
 				continue
--- a/cmd/ingest/ingest.go
+++ b/cmd/ingest/ingest.go
@@ -174,6 +174,7 @@ func main() {
 				}
 			}
 
+			dest.Close()
 			log.Debugln("Mark as archived")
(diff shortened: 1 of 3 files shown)


dtitov:
Why do we need goroutine?

jbygdell:
> Why do we need goroutine?

If we don't it will deadlock.

dtitov:
Can you elaborate a little bit more? I don't see any mutexes or semaphores here, so without a context it's not so obvious what and where can get deadlocked...

jbygdell:
If there is no goroutine it will not return to the main function as the uploader is locking until EOF
But as we are not returned to the main function we can't write any data to it and thus no EOF are recieved by the uploader.

jonandernovella:
> Can you elaborate a little bit more? I don't see any mutexes or semaphores here, so without a context it's not so obvious what and where can get deadlocked...

A pipe blocks on read or write until a corresponding write or read is called. Therefore a new goroutine is needed to write to the pipe and return it to be read from the main goroutine. That way we avoid the usage of unnecessary buffering.

jonandernovella:
> > Can you elaborate a little bit more? I don't see any mutexes or semaphores here, so without a context it's not so obvious what and where can get deadlocked...
> 
> A pipe blocks on read or write until a corresponding write or read is called. Therefore a new goroutine is needed to write to the pipe and return it to be read from the main goroutine. That way we avoid the usage of unnecessary buffering.

That is, you save a lot of memory :)

dtitov:
Ok, first of all, I don't think the method should be called `WriteFile` since it doesn't write anything: it creates and returns `Writer` (`WriteCloser`). Thus the name is quite confusing.

jbygdell:
> LGTM. POSIX should also be redone to use pipes.

Posix doesn't need to use pipes as there is no reader involved in writing data to a posix backend.

jbygdell:
> Ok, first of all, I don't think the method should be called `WriteFile` since it doesn't write anything: it creates and returns `Writer` (`WriteCloser`). Thus the name is quite confusing.

Rename it to `Filewriter`?

dtitov:
> Rename it to `Filewriter`?

Yes, `FileWriter` or maybe even `GetFileWriter` since it's a method and not a field.

dtitov:
Ok, after looking into S3 Go SDK I understood the problem: they don't provide `Writer`, they accept `Reader`, therefore we have to pipe it. LGTM now, just rename the method, please.

jbygdell:
> > Rename it to `Filewriter`?
> 
> Yes, `FileWriter` or maybe even `GetFileWriter` since it's a method and not a field.

`NewFileWriter` is actually more fitting

dtitov:
> `NewFileWriter` is actually more fitting
Agree, make it `NewFileWriter`.

PR #37, opened 2020-06-16
Use getObject instead of Downloader

jbygdell:
This speeds up reading data from S3 as we don't have to read the entire file at once


dtitov:
TBH, I've lost track of S3 code (don't have much time to thoroughly dive into it), so LGTM, but someone else should approve it.

jonandernovella:
I have tested it and works as expected.

Issue #30, opened 2020-06-12
Add mapper


jbygdell:
- [ ] Rework mapper to use our shared configuration
- [ ] Include in functionality tests

PR #38, opened 2020-06-16
Import mapper

jbygdell:
* Import mapper from sda-mapper
* implement our config schema
* uses new message format

Fixes: #30

### DIFF
--- a/cmd/mapper/mapper.go
+++ b/cmd/mapper/mapper.go
@@ -48,7 +48,7 @@ func main() {
 					// this should be handled by the SQL retry mechanism
 				}
 
-				err = db.MapfileToDataset(int(fileID), mappings.DatasetID)
+				err = db.MapFileToDataset(int(fileID), mappings.DatasetID)
 				if err != nil {
 					log.Errorf("MapfileToDataset failed, reason: %v", err)
 					// this should be handled by the SQL retry mechanism
--- a/cmd/mapper/mapper.go
+++ b/cmd/mapper/mapper.go
@@ -41,24 +41,16 @@ func main() {
 			if err != nil {
 				log.Errorf("Not a json message: %s", err)
 			}
-			for _, accessionID := range mappings.AccessionIDs {
-				fileID, err := db.GetFileIDByAccessionID(accessionID)
-				if err != nil {
-					log.Errorf("failed to get FileID from AccecssionID: %v", accessionID)
-					// this should be handled by the SQL retry mechanism
-				}
 
-				err = db.MapFileToDataset(int(fileID), mappings.DatasetID)
-				if err != nil {
-					log.Errorf("MapfileToDataset failed, reason: %v", err)
-					// this should be handled by the SQL retry mechanism
-				}
+			err = db.MapFilesToDataset(mappings.DatasetID, mappings.AccessionIDs)
+			if err != nil {
+				log.Errorf("MapfileToDataset failed, reason: %v", err)
+				// this should be handled by the SQL retry mechanism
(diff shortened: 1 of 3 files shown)


internal/postgres/postgres.go
```diff
@@ -172,6 +172,33 @@ func (dbs *SQLdb) MarkReady(accessionID, user, filepath, checksum string) error
...
+	if err != nil {
+		log.Errorf("something went wrong with the DB qurey: %s", err)
+		return 0, err
+	}
+
+	return fileID, nil
+}
+
+// MapfileToDataset maps a file to a dataset in the database
+func (dbs *SQLdb) MapfileToDataset(fileID int, datasetID string) error {
```
dtitov, line 189:
```suggestion
func (dbs *SQLdb) MapFileToDataset(fileID int, datasetID string) error {
```

cmd/mapper/mapper.go
```diff
@@ -0,0 +1,66 @@
...
+				log.Errorf("Not a json message: %s", err)
+			}
+			for _, accessionID := range mappings.AccessionIDs {
+				fileID, err := db.GetFileIDByAccessionID(accessionID)
+				if err != nil {
+					log.Errorf("failed to get FileID from AccecssionID: %v", accessionID)
+					// this should be handled by the SQL retry mechanism
+				}
+
+				err = db.MapfileToDataset(int(fileID), mappings.DatasetID)
```
dtitov, line 51:
```suggestion
				err = db.MapFileToDataset(int(fileID), mappings.DatasetID)
```

dtitov:
Also, in the `sda-mapper` I was using transactions so that either all files are mapped to the dataset or none. I don't see this transaction here. Shouldn't we have it?

jbygdell:
> So, the problem with the current DB schema is that one table needed for this functionality is available only for `lega_in` user and another only for `lega_out` user. That's why in the `sda-mapper` I had to provide two connection strings to have two connections. Is it supported in the current DB connectivity configuration?

The issue with that is not in this code but how the `lega_in` and  `lega_out` users are specified, which in my mind is completely broken. The proper way to fix that is to create a `lega_mapper` in the DB that can write to `local_ega_ebi` and read from  `local_ega`.

dtitov:
I agree with you. Just saying that if two separate connections are not supported in this code, then right now it simply won't work with the current schema (if we don't add this `lega_mapper` user to it).

jbygdell:
> I agree with you. Just saying that if two separate connections are not supported in this code, then right now it simply won't work with the current schema (if we don't add this `lega_mapper` user to it).

What really needs to be done is that each service should have a separate user in the DB and i this code is supposed to be `next generation` then why add the parts of the current setup that is completely broken to begin with.

dtitov:
I agree again :) But I think the approach to introduce the next-gen version that we agreed on was that first we make it work with the current schema (so that we can simply replace all the Python microservices with the Go microservices and make sure that everything works as before). And then, after that, we change the schema to a proper one and support these changes in the new code. 

If right now we will write the microservices for the new (non-existing yet) schema, then we will have no means to easily validate that everything is working as it used to...

jbygdell:
That decision has already been made for us, as we received a next gen message on our last ingest cycle attempt.

dtitov:
Yes, but MQ messages format is different from the DB schema: we will anyway have to adapt to the new messages, even in the current Python code we use, right? But it doesn't require us to change the schema in this particular moment.

jbygdell:
There is no issues with the schemas.

PR #41, opened 2020-06-18
Json validation

jbygdell:
* Perform validation of both incoming and outgoing messages
* Add JSON schema for the message between ingest and verify

Fixes: #36


dtitov:
Didn't you forget `go.sum`?

dtitov:
Also, I don't really get why would we validate ingestion-verification messages? We are sending these messages ourselves, don't we? So we can be sure that they are valid...

cmd/finalize/finalize.go
```diff
@@ -33,6 +36,8 @@ func main() {
 	defer mq.Connection.Close()
 	defer db.Close()
 
+	ingestAccession := gojsonschema.NewReferenceLoader("https://raw.githubusercontent.com/EGA-archive/LocalEGA/master/ingestion/schemas/ingestion-accession.json")
```
blankdots, line 39: I would create a local fallback to a schema and log that the schema there does not exist to avoid URL changing without notice

dtitov: Not only URL: content can be changed as well.

blankdots: true, so maybe a separate module for loading schemas and adding a fallback option inside e.g. erroring with specific message like schema content changed, url not accessible etc. ?

blankdots: or that is too complicated for this version ?

dtitov: I think it is. I would personally just dump their current schemas and would use local files in the validation. Apart from that, we can set up a completely separate cron-based job (maybe on Travis) to daily check that their schemas are untouched, so that we get notified if URLs/content changes.

cmd/verify/verify.go
```diff
@@ -15,23 +15,25 @@ import (
...
 	"github.com/elixir-oslo/crypt4gh/streaming"
+	"github.com/xeipuuv/gojsonschema"
 
 	log "github.com/sirupsen/logrus"
 )
 
 // Message struct that holds the json message data
 type Message struct {
-	Filepath     string `json:"filepath"`
+	Filepath     string `json:"filep_ath"`
```
jonandernovella, line 25: Typo?

pontus: Good catch, but this is moot anyway since it's filepath now.

PR #45, opened 2020-07-08
Feature/messaging

dbampalikis:
Contains fixes needed for messaging
- Typos in schemas
- Addition of files.verified queue, according to CEGA schema
- Typo in ingest.go
- Correct queues for verify and finalize

### DIFF
--- a/cmd/verify/verify.go
+++ b/cmd/verify/verify.go
@@ -22,22 +22,22 @@ import (
 
 // Message struct that holds the json message data
 type Message struct {
-	Filepath     string `json:"filepath"`
-	User         string `json:"user"`
-	FileID       int    `json:"file_id"`
-	ArchivePath  string `json:"archive_path"`
-	FileChecksum string `json:"encrypted_checksums"`
-	ReVerify     *bool  `json:"re_verify"`
+	Filepath           string      `json:"filepath"`
+	User               string      `json:"user"`
+	FileID             int         `json:"file_id"`
+	ArchivePath        string      `json:"archive_path"`
+	EncryptedChecksums []Checksums `json:"encrypted_checksums"`
+	ReVerify           *bool       `json:"re_verify"`
 }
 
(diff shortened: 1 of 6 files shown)


cmd/verify/verify.go
```diff
@@ -136,23 +136,24 @@ func main() {
...
-					// Send message to completed
-					c := Completed{
+					// Send message to verified
+					c := Verified{
 						User:     message.User,
 						Filepath: message.Filepath,
 						DecryptedChecksums: []Checksums{
-							{"SHA256", fmt.Sprintf("%x", hash.Sum(nil))},
+							{"sha256", fmt.Sprintf("%x", hash.Sum(nil))},
+							{"md5", "b60fa2486b121bed8d566bacec987e0d"},
```
blankdots, line 156: maybe for testing would be better to generate a random hash smth along the lines of:
```
import (
	"crypto/rand"
	"fmt"
)

func randToken() string {
	b := make([]byte, 16)
	rand.Read(b)
	return fmt.Sprintf("%x", b)
}
```
this will help us avoid any conflicts even though we are using the `sha256`

jonandernovella: Or perhaps even better a md5 hash. What's the point of the random hash @blankdots ?

blankdots: the point is mostly to avoid possible points of conflict anywhere in the flow (e.g. db, testing messages with cEGA), until we get the proper `md5` from the `crypt4gh` go library. Most likely there will be none but just a safety measure.

blankdots: but was just a suggestion, I already approve as is :slightly_smiling_face:

jonandernovella: Good, letä's merge then

jonandernovella:
@dbampalikis Pleease take a look at the linters. :-)

PR #53, opened 2020-08-06
Support taking extra parameters for location expansion

pontus:


pontus:
Current agreement is that we don't need this.

PR #60, opened 2020-08-10
Reduce memory consumption for verify

pontus:
Avoid reading data into memory to fead into Crypt4GHReader, use a special reader to
return a supplied header first and then return data read from a "proper" reader.

### DIFF
--- a/cmd/verify/verify.go
+++ b/cmd/verify/verify.go
@@ -138,14 +138,10 @@ func main() {
 				continue
 			}
 
-			var buf bytes.Buffer
-			buf.Write(header)
-			rw := io.ReadWriter(&buf)
-			if _, e := io.Copy(rw, f); e != nil {
-				log.Error(e)
-			}
+			hr := bytes.NewReader(header)
+			mr := io.MultiReader(hr, f)
 
-			c4ghr, err := streaming.NewCrypt4GHReader(rw, key, nil)
+			c4ghr, err := streaming.NewCrypt4GHReader(mr, key, nil)
 			if err != nil {
 				log.Error(err)
 				continue
@@ -155,6 +151,7 @@ func main() {
 			sha256hash := sha256.New()
 
 			stream := io.TeeReader(c4ghr, md5hash)
+
 			if _, err := io.Copy(sha256hash, stream); err != nil {
 				log.Error(err)
 				continue
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/aws/aws-sdk-go v1.34.7
-	github.com/elixir-oslo/crypt4gh v1.0.0
+	github.com/elixir-oslo/crypt4gh v1.0.1
 	github.com/google/uuid v1.1.1
 	github.com/lib/pq v1.8.0
 	github.com/sirupsen/logrus v1.6.0


pontus:
@dtitov pointed out that this should be possible with a multireader, will try a rewrite with that instead.

pontus:
Jocke reported issues on the stand up today, so we should not merge this now.

jonandernovella:
I will explore this and report at next stand up

jonandernovella:
Currently not working as the MultiReader is not compatible with the crypt4gh library.

dtitov:
It is compatible, here's the working gist: https://gist.github.com/dtitov/b110903e9f2690b32b1d4ed1251edbe0

`body.enc`: https://github.com/neicnordic/LocalEGA-DOA/raw/master/test/body.enc

My assumption is that you are not decoding header from HEX before feeding it to the reader...

jonandernovella:
@dtitov That is correct. So a standard reader does accept HEX, and the MultiReader does not?

dtitov:
No idea. I'm just saying that I've successfully decrypted 'header' + 'body' with Crypt4GH using `MultiReader`. I can try to look into this PR myself, to figure out what's happening here.

jonandernovella:
I will do a test with decoding the hex header, and report back

dtitov:
No need, actually: I've just checked the `db.GetHeader` method. It does the conversion from HEX under the hood. So this must be something else. I'll take a look.

jonandernovella:
> No need, actually: I've just checked the `db.GetHeader` method. It does the conversion from HEX under the hood. So this must be something else. I'll take a look.

Good catch. Let us know!

dtitov:
I rebased the branch to get both types of int-tests running: POSIX and S3. As you can see, POSIX tests are successful, the problem exists only for S3 tests. 

My new assumption is that the S3 reader implementation that we have currently, isn't compatible with `MultiReader`: it works only if we copy the data to the buffer first, but doesn't work if we read directly from it. 

@jbygdell, do you have any ideas, why does it happen?

pontus:
I had a thought (not verified as I have not yet figured out how to patch "external libraries" effectively) that the multireader implementation as is (possibly for s3 only) while working as per contract will possibly deliver less data then expected by the crypt4gh library, so I'd be interested in either a ReadFull instead at https://github.com/elixir-oslo/crypt4gh/blob/master/streaming/in.go#L135 or a check (possible loop) if it got less than the expected amount of data.

jbygdell:
> @jbygdell, do you have any ideas, why does it happen?

No, but the error message suggest that we try to decrypt an incomplete c4gh datablock.
might have to do with the fact that we are reading a tcp datastream.

dtitov:
@pontus, `ReadFull` is not applicable here, since there might fewer bytes than `encryptedSegmentBytes` in the `Reader` - and it's a normal case, because `encryptedSegmentBytes` is a maximum and not a precise size, so there might be less to read.

pontus:
Ok, then I don't understand what mechanism handles that case.

dtitov:
Do you mean the one that I described above? It's for the end of the file: the very last data segment may be smaller than the default segment size, thus we must not expect the segment to be always of the max length. And, if it's the case, we are trimming the segment to the actual amount of bytes read (line 144): `err := segment.UnmarshalBinary(encryptedSegmentBytes[:read])`

pontus:
If it's at end of data ReadFull would work as well (with the need to check for less than expected data)?

Sorry if that was unclear, I meant for a more general case, e.g. a reader returning a random number of bytes (less than the available space) mid-stream.

jbygdell:
Verified a 1GiB file, less than 64MiB RAM used by verify.

PR #63, opened 2020-08-11
Integration test

jbygdell:
Integration tests to be run on each PR or locally given the readme.


cmd/verify/verify.go
```diff
@@ -120,6 +121,14 @@ func main() {
 			}
 			keyFile.Close()
 
+			// fix to deal with a slow S3 metadata write,
```
dtitov, line 124: Can we have a retry loop here with some max timeout?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,106 @@
+name: integration
```
dtitov, line 1:
```suggestion
name: Integration tests
```

jbygdell: fixed

PR #64, opened 2020-08-11
Unittests for broker

blankdots:


pontus:
100% coverage at least. 

Noticed a few cases where I considered if one wanted to change the current behaviour without wanting to decide.

Also, two distinct approaches with a mock server originally brought in from s3-upload-proxy and a mocking of interfaces is now being used. I think that's okay for now, but one might want to unify that in the future.

internal/broker/broker_test.go
```diff
@@ -0,0 +1,327 @@
...
+	err = SendMessage(&b, "corrID1", "exchange", "routingkey", false, msg)
+	assert.Nil(t, err, "Unexpected error from SendMessage (not reliable)")
+
+	err = SendMessage(&b, "corrID1", "exchange", "routingkey", true, msg)
+	assert.Nil(t, err, "Unexpected error from SendMessage (reliable)")
+
+}
+
+var tMqconf = Mqconf{"127.0.0.1",
+	5672,
```
jbygdell, line 123:
```suggestion
	6565,
```

internal/broker/broker_test.go
```diff
@@ -0,0 +1,327 @@
...
+	return nil
+}
+
+// Test connection refused as there is no running
+// broker
+func TestNewMQConn_Error(t *testing.T) {
+
+	brokerURI := buildMqURI(tMqconf.Host, tMqconf.User, tMqconf.Password, tMqconf.Vhost, tMqconf.Port, tMqconf.Ssl)
+
+	expectedMsg := "dial tcp 127.0.0.1:5672: connect: connection refused"
```
jbygdell, line 221:
```suggestion
	expectedMsg := "dial tcp 127.0.0.1:6565: connect: connection refused"
```

jbygdell (changes requested):
Don't use real MQ ports as it will fail if MQ is running locally

jbygdell (approved):

PR #69, opened 2020-08-13
Tests/config

dtitov:
Here's the basic coverage of the `config` module. It covers all required settings and main errors, but I left out some more precise configuration like this:
```
	if viper.IsSet("broker.vhost") {
		if strings.HasPrefix(viper.GetString("broker.vhost"), "/") {
			b.Vhost = viper.GetString("broker.vhost")
		} else {
			b.Vhost = "/" + viper.GetString("broker.vhost")
		}
	} else {
		b.Vhost = "/"
	}
```


dtitov:
Ok, so somehow I've got here commits from other branches 😕 Not sure how to get rid of them...

dtitov:
Added more tests, 95% now.

jbygdell:
Added a few test, now we are at 100%

PR #75, opened 2020-08-18
Test try_decrypt from ingest

jbygdell:


cmd/ingest/ingest_test.go
```diff
@@ -0,0 +1,103 @@
...
+	viper.Set("c4gh.filepath", "/tmp/foo")
+	config, err := config.New("verify")
+	assert.NotNil(suite.T(), config)
+	assert.NoError(suite.T(), err)
+
+	buf := make([]byte, 1024)
+
+	byte, err := tryDecrypt(config.Crypt4gh, buf)
+	assert.Nil(suite.T(), byte)
+	assert.Error(suite.T(), err)
```
dtitov, line 53: I'd recommend asserting a specific error here (and in other places like this), to make sure that it's an error we are expecting and not some other.

cmd/ingest/ingest_test.go
```diff
@@ -0,0 +1,103 @@
...
+	buf := make([]byte, 65*1024)
+
+	file, err := os.Open("../../dev_utils/dummy_data.c4gh")
+	assert.NoError(suite.T(), err)
+
+	_, err = io.ReadFull(file, buf)
+	assert.NoError(suite.T(), err)
+
+	b, err := tryDecrypt(config.Crypt4gh, buf)
+	assert.NotNil(suite.T(), b)
```
dtitov, line 101: I'd also assert here the equality of the files (raw data and decrypted data). To make sure that the code around Crypt4GH doesn't add any "distortions" to the data. Otherwise, looks good.

PR #91, opened 2020-09-01
Intercept skeleton

jbygdell:
Simple service that parses messages coming from CRG based on the "type" field and passes them on to the correct internal routingkey.

### DIFF
--- a/cmd/intercept/intercept.go
+++ b/cmd/intercept/intercept.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"sda-pipeline/internal/broker"
+	"sda-pipeline/internal/config"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+func main() {
+	conf, err := config.NewConfig("intercept")
+	if err != nil {
+		log.Fatal(err)
+	}
+	mq, err := broker.NewMQ(conf.Broker)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	defer mq.Channel.Close()
+	defer mq.Connection.Close()
+
+	forever := make(chan bool)
+
+	log.Info("starting intercept service")
+
+	go func() {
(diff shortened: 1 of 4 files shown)


cmd/intercept/intercept.go
```diff
@@ -0,0 +1,108 @@
...
+			if err := delivered.Ack(false); err != nil {
+				log.Errorf("failed to ack message for reason: %v", err)
+			}
+		}
+	}()
+
+	<-forever
+}
+
+func validateJSON(body []byte) (string, *gojsonschema.Result, error) {
```
pontus, line 80: I'm unsure about how much heavy lifting the NewReferenceLoader does. Fine for now, but if we ever see visible resources usage from this, we might want to load these just once and reuse.

internal/config/config.go
```diff
@@ -100,24 +100,37 @@ func NewConfig(app string) (*Config, error) {
...
-		return nil, err
-	}
-
 	switch app {
 	case "ingest":
 		c.configInbox()
 		c.configArchive()
 		c.configCrypt4gh()
+		err = c.configDatabase()
+		if err != nil {
```
pontus, line 109: Not really to be addressed here, but we might want to consider if configDatabase should just give up and fatal if it can't get things up).

pontus (approved):

internal/config/config.go
jbygdell: These changes where introduced by Dmytro inorder to propagate the error back to the main function.

dtitov: Yep. I'm a bit hesitant about panicking, because it's usually hard to test. I prefer propagating errors upwards where possible.

cmd/intercept/intercept.go
dtitov: I don't foresee a high load in this place, we won't get so many messages so frequently. It's gonna be okay 👌

dtitov (approved):

internal/config/config.go
pontus: I kind of agree about the issues, but don't like the disharmony. But absolutely, this is fine for now.

PR #108, opened 2020-09-18
Bring over CA certificates from build image

pontus:


Dockerfile
```diff
@@ -22,5 +22,6 @@ LABEL org.label-schema.vcs-ref=$SOURCE_COMMIT
 COPY --from=builder /go/passwd /etc/passwd
 COPY --from=builder /go/sda-* /bin/
 COPY --from=builder /go/schemas /schemas
+COPY --from=builder /etc/ssl/certs/ /etc/ssl/certs/
```
jbygdell, line 25:
```suggestion
COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```

The only thing we need is the root ca file, everything else is just symlinks.

jbygdell (approved):

Issue #117, opened 2020-09-25
Memory stream-lining and concurrency 


pontus:
The AWS go sdk has some shenaigans and will be happy to try to use concurrency (and allocate buffers for this). To streamline memory usage, we should enable setting a lower concurrency than the default 5+1).

pontus:
This comes down to the maximum file size a deployment wishes to support and how that reflects on the partsize (governed by the maximum number of parts supported by an s3 backend, defaulting to 10 000).

Disabling concurrency would help to bring down memory consumption from partsize*10 to partsize*2. For a given memory limits, this enables support 5 times bigger partsizes (i.e. 5 times bigger files).

PR #118, opened 2020-09-25
Allow some concurrency handling

pontus:
One way of addressing #117 (that i'm not thoroughly happy with, especially falling back on memory allocation when concurrency is requested may be unhelpful and may need a better solution. We may also need to redo the tests to test with larger (multi-part) objects.

Actually make it possible to configure s3 concurrency.

Rewritten S3 object fetcher so it uses a downloader to enable passing a
concurrency.

Fallout from the rewritten object fetcher is revamped handling to
provider the reader.

### DIFF
--- a/internal/storage/storage.go
+++ b/internal/storage/storage.go
@@ -95,22 +95,22 @@ func (pb *posixBackend) GetFileSize(filePath string) (int64, error) {
 }
 
 type s3Backend struct {
-	Client   *s3.S3
-	Uploader *s3manager.Uploader
-	Bucket   string
+	Client     *s3.S3
+	Downloader *s3manager.Downloader
+	Uploader   *s3manager.Uploader
+	Bucket     string
 }
 
 // S3Conf stores information about the S3 storage backend
 type S3Conf struct {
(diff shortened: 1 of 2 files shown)


pontus:
It's quite possible that I've just missed a way to pass the Concurrency to an s3 client and most of the old code could be used, that would be better as far as I'm concerned.

pontus:
After some additional thinking, I think the best course of action is probably to disable any concurrency. I suspect that even if we assume speedup linear to concurrency, it will have no noticeable impact on the user experience.

jbygdell:
> After some additional thinking, I think the best course of action is probably to disable any concurrency. I suspect that even if we assume speedup linear to concurrency, it will have no noticeable impact on the user experience.

We also assume that the first data package we receive contains the header of the file.
If we use concurrent downloads we need to keep track of which part in the multipart series we are receiving.

pontus:
I think what provides the reader should be responsible for ordering data so reads get those in order. But I don't see us getting much benefit from the added complexity brought on by doing things simultaneously.

PR #129, opened 2020-10-05
Allow some concurrency handling

### LINKED ISSUE #117: Memory stream-lining and concurrency

pontus:
This is #118 as a zombie!

(This is intended to fix #117.)

While working on #127, I came up with this way that seems better. 

Actually make it possible to configure s3 concurrency.

Rewritten S3 object fetcher so it uses a downloader to enable passing a
concurrency.

To use the downloader with a pipe, we need an intermediate writer that supports the WriteAt interface. That writer keeps track of the amount of data it has written to not accept writes out of order.

To support concurrency, we use a lock to synchronize so access (both to the actual pipe and the current "offset" is available to just one thread at a time).

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -175,6 +175,10 @@ func configS3Storage(prefix string) storage.S3Conf {
 		s3.Chunksize = viper.GetInt(prefix+".chunksize") * 1024 * 1024
 	}
 
+	if viper.IsSet(prefix + ".concurrent") {
+		s3.Concurrency = viper.GetInt(prefix + ".concurrent")
+	}
+
 	if viper.IsSet(prefix + ".cacert") {
 		s3.Cacert = viper.GetString(prefix + ".cacert")
 	}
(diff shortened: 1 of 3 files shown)


jonandernovella:
As far as I know, the only conddition we need to fullfill is to read the first part's data block first, as it contains the key we need for unencryption. This seems to add quite some complexity and will make the code much harder to debug. I don't totally understand it yet.

pontus:
The "in order" part is always needed to not get data corrupted (but with the code currently in master, the AWS SDK handles that for us).

I don't think it's that complex, but would agree it would be best to avoid the issue, but due to the memory constraints as outlined in #117, I still think we need to do something.

One possible alternative is also to rely on the fact that DefaultDownloadConcurrency and DefaultUploadConcurrency are exported means they are stable enough to rely on and just set them.

pontus:
(Checked and I had misremembered, `DefaultDownloadConcurrency` and `DefaultUploadConcurrency` are const.

pontus:
Added so that the writing end closes the pipe when the expected amount of data has been transferred. That is helpful for ReadFull that will wait for new data to arrive otherwise.

Issue #112, opened 2020-09-21
Fail on startup if S3 backend can not be connected to


jbygdell:

jbygdell:
This was fixed in #136

Issue #111, opened 2020-09-21
Fail on startup if posix inbox/archive path does not exists


jbygdell:

pontus:
Fixed by #136.

PR #136, opened 2020-10-09
Bail out earlier on problems

### LINKED ISSUE #110: Fail on startup if c4gh keyfile can not be opended

pontus:
Moved some things that can be reused so they are initialized before the main loop. Also, verify storage backends upon creation.

Fixes #110, #111 and #112.

This currently uses ListObjectsV2 from s3 and requires that not to fail. That adds a requirement for s3 backends to have read permissions to the bucket. I believe that is acceptable (but worth mentioning especially).

### DIFF
--- a/cmd/ingest/ingest.go
+++ b/cmd/ingest/ingest.go
@@ -9,14 +9,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"os"
 
 	"sda-pipeline/internal/broker"
 	"sda-pipeline/internal/config"
 	"sda-pipeline/internal/database"
 	"sda-pipeline/internal/storage"
 
-	"github.com/elixir-oslo/crypt4gh/keys"
 	"github.com/elixir-oslo/crypt4gh/model/headers"
 	"github.com/elixir-oslo/crypt4gh/streaming"
 	"github.com/google/uuid"
@@ -61,7 +59,7 @@ func main() {
 		log.Fatal(err)
 	}
 
-	key, err := readKey(conf.Crypt4gh)
+	key, err := config.GetC4GHKey()
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -268,23 +266,6 @@ func main() {
 	}()
 
 	<-forever
-}
-
-// readKey reads and decrypts the c4gh key so it's ready for use
-func readKey(conf config.Crypt4gh) (*[32]byte, error) {
-	// Make sure the key path and passphrase is valid
-	keyFile, err := os.Open(conf.KeyPath)
-	if err != nil {
-		return nil, err
-	}
-
-	key, err := keys.ReadPrivateKey(keyFile, []byte(conf.Passphrase))
-	if err != nil {
-		return nil, err
-	}
-
-	keyFile.Close()
-	return &key, nil
 }
 
 // tryDecrypt tries to decrypt the start of buf.
--- a/cmd/ingest/ingest.go
+++ b/cmd/ingest/ingest.go
@@ -61,6 +61,23 @@ func main() {
 		log.Fatal(err)
 	}
 
+	key, err := readKey(conf.Crypt4gh)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	archive, err := storage.NewBackend(conf.Archive)
+	if err != nil {
+		log.Fatal(err)
+
+	}
+
+	inbox, err := storage.NewBackend(conf.Inbox)
+	if err != nil {
+		log.Fatal(err)
+
+	}
+
 	defer mq.Channel.Close()
 	defer mq.Connection.Close()
 	defer db.Close()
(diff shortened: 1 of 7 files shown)


cmd/verify/verify.go
```diff
@@ -214,3 +210,20 @@ func main() {
...
+	}
+
+	key, err := keys.ReadPrivateKey(keyFile, []byte(conf.Passphrase))
+	if err != nil {
+		return nil, err
+	}
+
+	keyFile.Close()
+	return &key, nil
+}
```
jbygdell, line 229: same here, better put it in config.go

internal/storage/storage.go
```diff
@@ -143,18 +165,30 @@ func newS3Backend(config S3Conf) *s3Backend {
...
+
+	return sb, nil
 }
 
 // NewFileReader returns an io.Reader instance
 func (sb *s3Backend) NewFileReader(filePath string) (io.ReadCloser, error) {
+	if sb == nil {
+		return nil, fmt.Errorf("Invalid s3Backend")
+	}
+
```
jbygdell, line 191: Is this necessary?An invalid back-end should have been caught at startup and prevented this point from being reached.

internal/storage/storage.go
```diff
@@ -170,6 +204,10 @@ func (sb *s3Backend) NewFileReader(filePath string) (io.ReadCloser, error) {
 
 // NewFileWriter uploads the contents of an io.Reader to a S3 bucket
 func (sb *s3Backend) NewFileWriter(filePath string) (io.WriteCloser, error) {
+	if sb == nil {
+		return nil, fmt.Errorf("Invalid s3Backend")
+	}
+
```
jbygdell, line 210: Is this necessary?An invalid back-end should have been caught at startup and prevented this point from being reached.

internal/storage/storage.go
```diff
@@ -187,6 +225,10 @@ func (sb *s3Backend) NewFileWriter(filePath string) (io.WriteCloser, error) {
 
 // GetFileSize returns the size of a specific object
 func (sb *s3Backend) GetFileSize(filePath string) (int64, error) {
+	if sb == nil {
+		return 0, fmt.Errorf("Invalid s3Backend")
+	}
+
```
jbygdell, line 231: Is this necessary?An invalid back-end should have been caught at startup and prevented this point from being reached.

jbygdell (changes requested):

cmd/verify/verify.go
pontus: Fixed similarly.

cmd/ingest/ingest.go
```diff
@@ -249,21 +263,29 @@ func main() {
...
+
+	key, err := keys.ReadPrivateKey(keyFile, []byte(conf.Passphrase))
 	if err != nil {
-		log.Error(err)
 		return nil, err
 	}
+
+	keyFile.Close()
+	return &key, nil
+}
```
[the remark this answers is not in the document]
pontus: Yep, mostly went for minimal refactoring. Moved to config now (would also be happy to but in a separate package, but either's fine).

internal/storage/storage.go
pontus: Should (of course) not be needed, no. With that said, these are cheap in the normal case and simplifies test writing a bit (or rather understanding if things fail).

My understanding of what happened is that a later crash made the prior assert not show so it took longer than it should have to figure out what happened.

I have a slight preference for leaving them because of that but am totally fine with pulling them out if you don't like them.

jonandernovella: As far as I know, it is a good practice to have this. It is not a crucial component, but still makes sure that you can track down the error better.

cmd/ingest/ingest.go
```diff
@@ -61,6 +59,23 @@ func main() {
 		log.Fatal(err)
 	}
 
+	key, err := config.GetC4GHKey()
+	if err != nil {
+		log.Fatal(err)
+	}
```
jbygdell, line 65: Putting the comment here as the code is spread out a bit.

The reason we didn't do it this way initially was to mitigate one of the flaws in the python version, that the key is always unlocked when the app is running.

That's why I suggested moving the unlocking of the key to the `config` part, since it is used by more than one component, while still returning the keypath and the passphrase to the app. That way we test that the key can be unlocked and we fail early if it can't. But we only unlock it when it is time to use it.

pontus: I don't have a strong opinion on keeping a single key fetch, but I don't see that helping security much (for most attacks, it's likely that the attacker has access to the file and can more easily grab the passphrase from the environment than analyze a memory dump to try to locate something that looks like a key).

If we move it back inside the loop, we might as well clear the array used for the key when we're done with it.

jbygdell:
Needs to be rebased ontop of master.

jbygdell (approved):

blankdots (approved):

PR #139, opened 2020-10-14
for standalone version we should accept any id

### LINKED ISSUE #138: allow custom accession id patterns for standalone deployment
current deployment is restricted to ^EGAF[0-9]{11}$'] pattern however to use it in standalone mode one would need to enable a custom pattern.

The aim is to enable custom schema depending on deployment

blankdots:
closes #138 
accept anything for accession ids
as urn and DOIs can contain special identifiers this cannot create a regex for either one that will exclude them.


pontus (approved):
I'd possibly consider disallowing the empty string as a valid accession id, but have no really strong opinion on the matter.

blankdots:
@pontus good point ... done.

jbygdell (commented):
Maybe we should be a bit more restrictive that this,
We should make sure that the string doesn't contain special characters.

blankdots:
@jbygdell true, but I will make a config switch as discussed in chat and refine the schema a for special characters

pontus:
(I  rebased.)

pontus (approved):

PR #141, opened 2020-10-15
fix to correct archive filesize

blankdots:


cmd/ingest/ingest.go
```diff
@@ -201,7 +201,7 @@ func main() {
 			log.Debugln("Mark as archived")
 			fileInfo := database.FileInfo{}
 			fileInfo.Path = archivedFile
-			fileInfo.Size = fileSize
+			fileInfo.Size, err := archive.GetFileSize(archivedFile)
```
jbygdell, line 204:
```suggestion
			fileInfo.Size, err := archive.GetFileSize(archivedFile)
			if err != nil {
				log.Errorf("Failed to get file size of: %s, reason: %v", archivedFile, err)
				continue
			}
```

jbygdell (changes requested):
Need to catch the error

cmd/ingest/ingest.go
blankdots: my editor complains about `:=` so `=` seems to work

blankdots:
I would approve but I cannot as i started this, so I think the changes are ok and we can merge.

pontus:
I think only admin can approve as long as anyone (e.g. @jbygdell) has requested changes.

jbygdell (approved):

PR #144, opened 2020-10-15
Better logs from tests

pontus:
Show logs from earlier services if a service fail.


blankdots:
I  think we can close this

pontus:
I would like to fix this so it gets a bit easier to see the reason for failures when they appear, don't have a timeline for that, though.

blankdots:
oh ok, then we can keep.

blankdots (approved):

jbygdell (approved):

PR #148, opened 2020-10-20
Mq updates

jbygdell:
Updates related to the changes in our MQ layout.
Adds both mapper and intercept to the tests
Includes a bugfix that prevented an error from being propagated properly.


.github/integration/tests/common/30_ingest_test.sh
```diff
@@ -123,5 +122,40 @@ for file in dummy_data.c4gh largefile.c4gh; do
...
+                           "payload_encoding":"string",
+                           "payload":"{
+                              \"type\":\"mapping\",
+                              \"dataset_id\":\"DATASET\",
+                              \"accession_ids\":[\"ACCESSIONID\"]}"
+                           }'| sed -e "s/DATASET/$dataset/" -e "s/ACCESSIONID/$access/" -e "s/CORRID/$count/")"
+
+   dbcheck=$(docker run --rm --name client --network dev_utils_default \
+   neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
+   -c "SELECT * from local_ega_ebi.file_dataset where dataset_id='$dataset' and file_id='$access'")
```
pontus, line 149: This will have output with column headers and thus always have a length, a `-t` should fix that.

pontus (approved):

.github/integration/tests/common/30_ingest_test.sh
jbygdell: When you get a response yes, if there is no match to the query you get a null response.

blankdots (approved):
all good from my point of view.

PR #153, opened 2020-10-22
Store decrypted file size (and other attributes).

### LINKED ISSUE #142: calculate un-encrypted file size in `verify`
we should calculate un-encrypted file size in `verify`

Problem is which column to use to store in db. `inbox_filesize` seem unnecessary at the moment and could be used

pontus:
Fixes #142.

Try to clean up some confusing naming issues as well - the database field `archive_file_checksum` now stores the checksum of what is actually stored in the archive rather than the decrypted checksum.

Decrypted checksum is stored in `decrypted_file_checksum`.

Sizes of archived and decrypted files are also kept (`archive_file_size` and `decrypted_file_size`).

### DIFF
--- a/.github/integration/tests/common/30_ingest_test.sh
+++ b/.github/integration/tests/common/30_ingest_test.sh
@@ -266,7 +266,7 @@ done
 
 docker run --rm --name client --network dev_utils_default \
 neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
--t -c "SELECT * from local_ega_ebi.file_dataset ORDER BY id DESC"
+-t -c "SELECT * from local_ega_ebi.file_dataset"
 
 docker run --rm --name client --network dev_utils_default \
 neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
--- a/.github/integration/tests/common/30_ingest_test.sh
+++ b/.github/integration/tests/common/30_ingest_test.sh
@@ -2,7 +2,7 @@
 
 docker run --rm --name client --network dev_utils_default \
 neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
--t -c "SELECT * from local_ega_ebi.file_dataset ORDER BY id DESC"
+-t -c "SELECT * from local_ega_ebi.file_dataset"
 
 docker run --rm --name client --network dev_utils_default \
 neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
--- a/.github/integration/tests/common/30_ingest_test.sh
+++ b/.github/integration/tests/common/30_ingest_test.sh
@@ -31,9 +31,14 @@ for file in dummy_data.c4gh largefile.c4gh; do
     C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | sed -e 's/.* //' -e 's/"//g')
     export C4GH_PASSPHRASE
 
-    decsha256sum=$(crypt4gh decrypt --sk c4gh.sec.pem < "$file" | sha256sum | cut -d' ' -f 1)
+    dcf=$(tempfile)
+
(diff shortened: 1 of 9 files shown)


internal/database/db.go
```diff
@@ -44,14 +45,22 @@ type DBConf struct {
...
+	DecryptedChecksum hash.Hash
+	DecryptedSize     int64
 }
 
 // Internal variable to ease testing
 var sqlOpen = sql.Open
 
+// hashType returns the identification string for the hash type
+func hashType(h hash.Hash) string {
+	// TODO: Support/check type
```
blankdots, line 60: should we add an issue about this, to keep better track of it ?

pontus: :+1:

pontus:
(I'm not sure why the integration actually passes for this, will fix and reopen.)

blankdots (approved):
LGTM

.github/integration/tests/common/30_ingest_test.sh
```diff
@@ -2,7 +2,7 @@
 
 docker run --rm --name client --network dev_utils_default \
 neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
--t -c "SELECT * from local_ega_ebi.file_dataset ORDER BY id DESC"
+-t -c "SELECT * from local_ega_ebi.file_dataset"
```
jbygdell, line 5: Why remove the sorting?
With the `10_trigger_failures` script running it is nice to see that it is really the last two inserted files that gets the the Accession and dataset ids

.github/integration/tests/common/30_ingest_test.sh
```diff
@@ -261,12 +266,38 @@ for file in dummy_data.c4gh largefile.c4gh; do
...
+      #exit 1
+   fi
+
     count=$((count+1))
 done
 
 docker run --rm --name client --network dev_utils_default \
 neicnordic/pg-client:latest postgresql://lega_out:lega_out@db:5432/lega \
--t -c "SELECT * from local_ega_ebi.file_dataset ORDER BY id DESC"
+-t -c "SELECT * from local_ega_ebi.file_dataset"
```
jbygdell, line 300: Same here why remove the sorting

cmd/verify/verify.go
```diff
@@ -146,18 +158,24 @@ func main() {
...
 				continue
 			}
 
+			file.Checksum = archiveFileHash
+			file.DecryptedChecksum = sha256hash
+
 			//nolint:nestif
 			if message.ReVerify == nil || !*message.ReVerify {
-				log.Debug("Mark completed")
+				log.Debug("will run markcompleted")
```
jbygdell, line 171: This seems redundant

.github/integration/tests/common/30_ingest_test.sh
pontus: The view local_ega_ebi.file_dataset does not have any `id` column, if desirable we can sort on one of the ids available.

.github/integration/tests/common/30_ingest_test.sh
pontus: Same as above.

.github/integration/tests/common/30_ingest_test.sh
```diff
@@ -283,7 +283,7 @@ for file in dummy_data.c4gh largefile.c4gh; do
 		      neicnordic/pg-client:latest postgresql://lega_in:lega_in@db:5432/lega \
 		      -t -c "SELECT decrypted_file_checksum from local_ega.files where stable_id='$access';")
 
-   if [ "$decryptedchecksum" = "$decsha256sum" ]; then
+   if [ "$(echo $decryptedchecksum)" = "$decsha256sum" ]; then
```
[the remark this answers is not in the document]
pontus: Changed to use that in 9ae6ccc2b2df6b189a078dcfe8ea2e84ff00e4d0.

cmd/verify/verify.go
pontus: Addressed in 846fb9a0ba5043b88fa5486d55b38c2419c0295e.

PR #156, opened 2020-10-26
Mq detect connection failure

### LINKED ISSUE #152: sda-pipeline reconnect to MQ

jbygdell:
Simplest solution to the problem of MQ re-connectivity
* Detect if the MQ connection closes and exit the main process.
* Rely on the deployment system to restart a stopped application.

fixes #152 

If we want to keep the main  process running a major rewrite is required.

### DIFF
--- a/cmd/finalize/finalize.go
+++ b/cmd/finalize/finalize.go
@@ -5,6 +5,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 
 	"sda-pipeline/internal/broker"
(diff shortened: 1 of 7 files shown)


pontus:
How about moving the actual gofunc to broker and just having a call to signal the addition of a watcher?

blankdots (approved):
fine with also what @pontus proposed but also ok with how it is.

if we want to have the auto-reconnect at some point maybe we should do a follow-up issue

pontus (approved):

PR #161, opened 2020-10-27
filepath from CRG means inbox filepath

jbygdell:


blankdots (approved):
I approve as I know this will be fixed once the integration tests are fixed

pontus (approved):

PR #166, opened 2020-10-29
Mq messages on error

### LINKED ISSUE #130: Send messages on error
Upon failures that aren't likely local/temporary, we need to make sure we send a message to the correct queue for errors queue for communication to the user as well as nacking the message so we do not get our queues filled up with messages that only lead to failures.

jbygdell:
* Moves JSON validation to a separate function
* Added test for the above function
* Send messges on JSON error
* Send messages on file related errors for ingest and verify

fixes #130

### DIFF
--- a/cmd/finalize/finalize.go
+++ b/cmd/finalize/finalize.go
@@ -15,27 +15,27 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// Message struct that holds the json message data
-type Message struct {
+// finalize struct that holds the json message data
+type finalize struct {
 	Type               string      `json:"type"`
 	User               string      `json:"user"`
 	Filepath           string      `json:"filepath"`
 	AccessionID        string      `json:"accession_id"`
-	DecryptedChecksums []Checksums `json:"decrypted_checksums"`
+	DecryptedChecksums []checksums `json:"decrypted_checksums"`
 }
 
 // Checksums is struct for the checksum type and value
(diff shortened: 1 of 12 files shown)


cmd/finalize/finalize.go
```diff
@@ -79,61 +77,124 @@ func main() {
...
+				josnError := broker.JSONError{
+					Error:          "Handling json message failed",
+					Reason:         err.Error(),
+					OrginalMessage: delivered.Body,
+				}
+				body, _ := json.Marshal(josnError)
+
+				if e := broker.SendMessage(mq, delivered.CorrelationId, conf.Broker.Exchange, conf.Broker.RoutingError, conf.Broker.Durable, body); e != nil {
+					log.Error("faild to publish message, reason: ", e)
+				}
```
pontus, line 160: Can't this be made a function in broker rather than having this much code for messages everywhere)?

jbygdell: I was thinking about it

pontus (approved):
I haven't actually verified but seems fine in principle, though I think it would very good to bring as much as possible into broker to lower code duplication._

blankdots (commented):
small typo `errorus` should be `erroneous `

`// TODO fix resend mechanism` - an issue about this would track it better

blankdots (approved):
same comment as @pontus about code duplication

blankdots (approved):

PR #175, opened 2020-11-06
Add db-retry mechanism

pontus:
Retry database operation a configurable number of times.

Make operations idempotent (or close enough), specifically
MapFilesToDataset.

Also only transition files with status COMPLETED in MarkReady.

### DIFF
--- a/internal/database/db.go
+++ b/internal/database/db.go
@@ -137,9 +137,13 @@ func (dbs *SQLdb) checkAndReconnectIfNeeded() {
 
 // GetHeader retrieves the file header
 func (dbs *SQLdb) GetHeader(fileID int) ([]byte, error) {
-	r, err, count := nil, nil, 0
+	var (
+		r     []byte = nil
+		err   error  = nil
+		count int    = 0
+	)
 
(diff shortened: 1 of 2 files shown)


internal/database/db.go
```diff
@@ -137,20 +137,30 @@ func (dbs *SQLdb) checkAndReconnectIfNeeded() {
 
 // GetHeader retrieves the file header
 func (dbs *SQLdb) GetHeader(fileID int) ([]byte, error) {
+	r, err := dbs.getHeader(fileID)
+	count := 0
+
+	for err != nil && count < dbRetryTimes {
+		r, err = dbs.getHeader(fileID)
+		count++
+	}
```
jbygdell, line 146: I would have done it like this to get rid of the first DB call.

```suggestion
	r, err, count := nil, nil, 0

	for ( r == nil || err != nil) && count < dbRetryTimes {
		r, err = dbs.getHeader(fileID)
		count++
	}
```

jbygdell (approved):
LGTM, although I would have done it slightly different as per comment.

blankdots (approved):
LGTM, looks like a typing issue that will most likely be fixed, so pre-approving

PR #182, opened 2020-11-13
Some log cleanup and make it easier to follow flow

pontus:


jbygdell (approved):

jonandernovella (changes requested):
Looks good overall! Though I think we better do this using the logger's native features?

E.g: `log.WithFields(log.Fields{"corrId": "ega001"}).Errorf("Copy failed while reading file: %v", err))`

PR #189, opened 2020-11-19
Initial implementation of syncing

dbampalikis:


cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+			log.Debug("Mark ready")
+			if err := db.MarkReady(message.AccessionID, message.User, message.Filepath, checksumSha256); err != nil {
+				log.Errorf("MarkReady failed, reason: %v", err)
+				// nack the message but requeue until we fixed the SQL retry.
+				if e := delivered.Nack(false, true); e != nil {
+					log.Errorln("failed to Nack message, reason: ", e)
+				}
+				continue
+				// this should be handled by the SQL retry mechanism
+			}
```
pontus, line 199: This belongs to finalize?

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+					log.Errorln("failed to Nack message, reason: ", e)
+				}
+
+				// Send the message to an error queue so it can be analyzed.
+				if e := mq.SendJSONError(&delivered, err.Error(), conf.Broker); e != nil {
+					log.Error("failed to publish message, reason: ", res.Errors())
+				}
+				// Restart on new message
+				continue
+			}
```
pontus, line 117: The logging additions we merged recently also includes moving roughly this to broker, so this can be shortened to a single call.

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+				if e := delivered.Nack(false, false); e != nil {
+					log.Errorln("failed to Nack message, reason: ", e)
+				}
+				// Send the message to an error queue so it can be analyzed.
+				if e := mq.SendJSONError(&delivered, err.Error(), conf.Broker); e != nil {
+					log.Error("failed to publish message, reason: ", err)
+				}
+				// Restart on new message
+				continue
+			}
```
pontus, line 166: This can also be turned into a single call with handling.

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+			dest, err := backup.NewFileWriter(message.Filepath)
+			if err != nil {
+				log.Errorf("Failed to create file, reason: %v", err)
+				continue
+			}
+
+			_, err = io.Copy(dest, file)
+			if err != nil {
+				log.Fatal(err)
+			}
```
pontus, line 184: I don't see any upside to doing a Fatal here. Not sure what we want to do though (requeue? send to error queue?)

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+			// we unmarshal the message in the validation step so this is safe to do
+			_ = json.Unmarshal(delivered.Body, &message)
+
+			// Extract the sha256 from the message and use it for the database
+			var checksumSha256 string
+			for _, checksum := range message.DecryptedChecksums {
+				if checksum.Type == "sha256" {
+					checksumSha256 = checksum.Value
+				}
+			}
```
pontus, line 128: If we leave `MarkReady` in `finalize`, we don't  need `checksumSha256`.

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+	db, err := database.NewDB(conf.Database)
+	if err != nil {
+		log.Fatal(err)
+	}
+	backup, err := storage.NewBackend(conf.Backup)
+	if err != nil {
+		log.Fatal(err)
 
+	}
 	inbox, err := storage.NewBackend(conf.Inbox)
```
pontus, line 60: This should be `Archive` rather?

cmd/sync/sync.go
dbampalikis: The idea is to deploy either sync or finalize. So I just included the sync in the already existing code of finalize

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
+				continue
+			}
+
+			dest, err := backup.NewFileWriter(message.Filepath)
+			if err != nil {
+				log.Errorf("Failed to create file, reason: %v", err)
+				continue
+			}
+
+			_, err = io.Copy(dest, file)
```
pontus, line 181: Probably overkill, but we could consider adding ability to fetch archived size from database and verify that with the number of bytes copied.

internal/config/config.go
```diff
@@ -133,6 +140,15 @@ func NewConfig(app string) (*Config, error) {
 		}
 		return c, nil
 	case "finalize":
+		err = c.configDatabase()
+		if err != nil {
+			return nil, err
+		}
+		return c, nil
+	case "sync":
+		c.configInbox()
```
pontus, line 149: Archive.

cmd/sync/sync.go
```diff
@@ -3,64 +3,232 @@
...
-	<-forever
+	_, ok := message["type"]
+	if ok {
+		schema = gojsonschema.NewReferenceLoader(schemasPath + "ingestion-accession.json")
+	} else {
+		schema = gojsonschema.NewReferenceLoader(schemasPath + "ingestion-completion.json")
+	}
+	res, err := gojsonschema.Validate(schema, gojsonschema.NewBytesLoader(body))
+	return res, err
 }
```
pontus: Moved to broker since recent merge.

cmd/sync/sync.go
pontus: Ah, that wasn't my understanding but in that case I guess it's fine.

cmd/sync/sync.go
dbampalikis: I am not sure about that. Maybe @jbygdell has some input?

jbygdell: Should be archive or the bucket and access keys will be wrong.

For this service `archive` is the source and `backup` the destination

dev_utils/compose-sda.yml
```diff
@@ -48,6 +48,19 @@ services:
...
+    depends_on: 
+      - certfixer
+    env_file: ./env.sync
+    image: neicnordic/sda-pipeline:latest
+    volumes:
+      - ./config.yaml:/config.yaml
+      - ./:/dev_utils/
+      - certs:/dev_utils/certs
+    mem_limit: 64m
+    restart: always
```
jbygdell, line 63: Wont this create a round robin scenario since both finalize and sync reacts to the same messages?

dev_utils/env.sync
```diff
@@ -0,0 +1,6 @@
+BROKER_EXCHANGE=sda
+BROKER_HOST=mq
+BROKER_QUEUE=accessionIDs
+BROKER_ROUTINGKEY=completed
+BROKER_ROUTINGERROR=error
+DB_HOST=db
```
jbygdell, line 6: Missing archive and backup config

cmd/sync/sync.go
jbygdell: If we get an error here we should requeue, log to error and take the next message

jbygdell (changes requested):
Needs to be rebased on master to get the latest changes for logging, message validation and so.

dev_utils/compose-sda.yml
dbampalikis: But aren't we supposed to have only sync or finalize, but not both?

jbygdell: Yes, but what you have here will start both if you do `docker-compose up`.
It will work but we need to document that the unwanted service needs to be shut down before any tests are run.

pontus: Better to have them use different queues and configure `verify` to send to the desired one?

pontus: (From principle, while it's quite minimal, I'd rather prefer not having the functionality duplicated but rather have `sync` act as a pass-through, sending onwards to `finalize` once the copy is done. Both alternatives obviously works, though.)

jbygdell:
Closed, superseded by #214

jbygdell:
Closed, superseded by #214

PR #194, opened 2020-11-24
update actions go version

blankdots:
use go 1.14 and 1.15 to test
bump actions version for python and go
cleanup comments

### DIFF
--- a/.github/workflows/create_docs.yaml
+++ b/.github/workflows/create_docs.yaml
@@ -21,7 +21,7 @@ jobs:
           ref: gh-pages
 
       - name: Set up Go 1.14
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v2
         with:
           go-version: 1.14
         id: go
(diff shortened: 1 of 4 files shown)


.github/workflows/create_docs.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
           ref: gh-pages
 
       - name: Set up Go 1.14
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v2
         with:
           go-version: 1.14
```
jonandernovella, line 26: `1.15` should be there too?

jonandernovella (changes requested):

.github/workflows/create_docs.yaml
blankdots: that creates docs, so i think 1 version is enough, i can put the latest

pontus (approved):

jbygdell (approved):

.github/workflows/create_docs.yaml
blankdots: i pushed the latest @jonandernovella is that ok or should i still go with 2 versions ?

jonandernovella (approved):

PR #195, opened 2020-11-24
use sda-mq queue names

blankdots:
these files are not required for anything it seems


pontus (approved):

jbygdell (approved):
They where used before the changes to sda-mq had been merged.

PR #197, opened 2020-11-27
Move to a unified ValidateJSON and fall-out from there

pontus:
Validation/JSON handling moved to broker package.

The unmarshalling test is stricter now and will complain if fields
from the JSON is missing in the destination struct.

(I'm ambivalent as to whatever having JSON handling in broker or a new package, so feel free to say if you think it's better in a new package.)


jbygdell (approved):
LGTM

PR #199, opened 2020-11-30
fix logs exposing information, according to codeql

blankdots:


internal/database/db.go
```diff
@@ -129,7 +129,7 @@ func (dbs *SQLdb) checkAndReconnectIfNeeded() {
 			logFatalf("Could not reconnect to failed database in reasonable time, giving up")
 		}
 		time.Sleep(dbReconnectSleep)
-		log.Debugf("Reconnecting to DB with <%s>", dbs.ConnInfo)
+		log.Debugf("Reconnecting to DB with")
```
jbygdell, line 132:
```suggestion
		log.Debugln("Reconnecting to DB")
```
The `with` at then end felt a bit dangling.

jbygdell (approved):

pontus (approved):

PR #224, opened 2021-01-14
Include sync in integration tests

pontus:


cmd/intercept/intercept.go
```diff
@@ -89,14 +89,18 @@ func main() {
 			}
 
 			routing := map[string]string{
-				msgAccession: "accessionIDs",
+				msgAccession: conf.Route.RouteAccessionIDs,
```
jbygdell, line 92:
```suggestion
				msgAccession: "accessionIDs",
```

dev_utils/env.intercept
```diff
@@ -3,4 +3,5 @@ BROKER_HOST=mq
 BROKER_QUEUE=files
 BROKER_ROUTINGKEY=ingest
 BROKER_ROUTINGERROR=error
+ROUTE_ACCESSIONIDS=backup
```
jbygdell, line 6:
```suggestion
```

internal/config/config.go
```diff
@@ -24,13 +24,18 @@ const S3 = "s3"
 
 var requiredConfVars []string
 
+type RouteConfig struct {
+	RouteAccessionIDs string
+}
+
```
jbygdell, line 30:
```suggestion
```

internal/config/config.go
```diff
@@ -24,13 +24,18 @@ const S3 = "s3"
...
 // Config is a parent object for all the different configuration parts
 type Config struct {
 	Archive  storage.Conf
 	Broker   broker.MQConf
 	Inbox    storage.Conf
 	Backup   storage.Conf
 	Database database.DBConf
+	Route 	 RouteConfig
```
jbygdell, line 38:
```suggestion
```

internal/config/config.go
```diff
@@ -129,6 +134,7 @@ func NewConfig(app string) (*Config, error) {
 		}
 		return c, nil
 	case "intercept":
+		c.configIntercept()
```
jbygdell, line 137:
```suggestion
```

internal/config/config.go
```diff
@@ -303,6 +309,15 @@ func (c *Config) configBroker() error {
...
 
+// configIntercept provides configuration for intercept
+func (c *Config) configIntercept() {
+	r := RouteConfig{}
+
+	viper.SetDefault("route.accessionids", "accessionIDs")
+	r.RouteAccessionIDs = viper.GetString("route.accessionids")
+	c.Route = r
+}
+
```
jbygdell, line 320:
```suggestion
```

internal/config/config_test.go
```diff
@@ -465,6 +465,16 @@ func (suite *TestSuite) TestInterceptConfiguration() {
...
+func (suite *TestSuite) TestInterceptDefault() {
+	config, err := NewConfig("intercept")
+	assert.NoError(suite.T(), err)
+	assert.Equal(suite.T(), config.Route.RouteAccessionIDs, "accessionIDs")
+	viper.Set("route.accessionids", "testvalue")
+	config, err = NewConfig("intercept")
+	assert.NoError(suite.T(), err)
+	assert.Equal(suite.T(), config.Route.RouteAccessionIDs, "testvalue")
+}
+
```
jbygdell, line 477:
```suggestion
```

jbygdell (changes requested):
Changes to `intercept` should not be necessary, it is designed to be a transparent part only used in federated mode.

pontus:
Not sure what happened, I thought I wrote a description here.

The alternative to having a configurable routing for accessionID that I see for supporting having finalize and sync in different orders is to handle it via queue names, which felt more abusive/confusing.

jbygdell:
The biggest issue I see here is that these changes have no effect in a standalone setup, where intercept is not used.

pontus:
Correct, in a standalone setup, I don't see the same issues (I don't  mind e.g. having the orchestrator send to the sync/backup queue).

blankdots:
but even in federated setup still no issue as the `backup` queue is internal so it does not need to pass through intercept `sync` will only send to `completed`. or is there an issue i am missing ?

jbygdell:
No you are not, hence my reluctance to involve `intercept` in this.

blankdots:
then let us not involve `intercept` and just focus on `sync` functionality in `s3` and `posix` by the easiest way there is.

*If* we need to involve `intercept` i assume it is a different use case, different test and different PR.

pontus:
The issue is if one wants to have non-locked order of the services and don't want to use queues in a different order than we  otherwise do:

intercept is hardcoded to route incoming accessionID messages to the queue accessionIDs.

So if one wants to run sync before finalize, sync will need to listen to accessionIDs leaving finalize to listen to the backup queue. 

The mixing up of naming there is the part I think is less than optimal. 

I've created #229 which I hope can be merged.

pontus:
Closing this without deleting the branch for now.

PR #238, opened 2021-02-03
Add missing config tests

jbygdell:


internal/config/config_test.go
```diff
@@ -188,6 +223,52 @@ func (suite *TestSuite) TestConfigS3Storage() {
 	assert.Equal(suite.T(), "test", config.Archive.S3.Cacert)
 }
 
+func (suite *TestSuite) TestConfigSyncS3Storage() {
+	viper.Set("archive.type", S3)
+	viper.Set("archive.url", "test")
+	viper.Set("archive.accesskey", "test")
+	viper.Set("archive.secretkey", "test")
```
pontus, line 230: Possibly set different values for different fields so we verify that we pick the correct value to the correct field.

pontus (approved):

blankdots (approved):

PR #310, opened 2021-11-10
removing unused db interface

blankdots:
that seems it is not used, shall we remove it ?


blankdots:
we can reopen when if we decide to remove it.

PR #312, opened 2021-11-17
[HotFix] for stalled ingestions

jbygdell:
By moving the creation of the NotifyPublish channel to the main MQ channel creation, instead of message publishing a deadlock is prevented. For some reason the channel was not cleared by listening on the response.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -236,19 +236,6 @@ func TLSConfigBroker(config MQConf) (*tls.Config, error) {
 	return &tlsConfig, nil
 }
 
-// confirmOne accepts confirmation for a message sent with reliable.
-//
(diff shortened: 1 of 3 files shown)


blankdots (approved):

dbampalikis (approved):

blankdots:
we should start tagging versions on master i see we have `0.1.1` ... maybe this can be `0.1.2`

jbygdell:
> we should start tagging versions on master i see we have `0.1.1` ... maybe this can be `0.1.2`

Yes, I have found an Action that will do versioning on each PR, that way we won't forget it.

norling (approved):
Nice! There was a typo on line 97 of `broker_test.go`, but that doesn't seem critical :)

PR #318, opened 2021-11-29
Refactor/repo cleanup

blankdots:
- [x] fix typos in the code and docs
- [x] remove docker hub hooks
- [x] adjust documentation to make it somewhat more understandable

Question: do we active `deepsource` ? we might need to fix issues if we do

This will go as `patch` version update as we made some code changes.

PR #319, opened 2021-12-01
Reduce fetch depth #none

jbygdell:
Reduce fetch depth to 1 in order to prevent issues with new tags


blankdots (approved):
i think we might need to fix mq image as well till we have no tls helm support as with PR 
#318

jbygdell:
> i think we might need to fix mq image as well till we have no tls helm support as with PR #318

I can look into that in a separate PR

PR #328, opened 2021-12-09
Feature/add shellcheck action #none

### LINKED ISSUE #322: Add shellsheck and shfmt to actions suite
Run on each push to any branch except master.

blankdots:
Closes #322 
changes:
- adjust shell scripts with a uniform syntax
- add shellcheck github action that runs only on all branches except master
- don't require python 3.6 as default version in integration tests
- tag whole team for checking dependency updates

### DIFF
--- a/.github/integration/setup/common/10_services.sh
+++ b/.github/integration/setup/common/10_services.sh
@@ -14,7 +14,7 @@ if [ "$STORAGETYPE" = s3 ]; then
     tostart="mq db s3 certfixer"
 fi
 
-# we need to let it be split here otherwise it takes it as one
+# We need to leave the $tostart variable unquoted here since we want it to split
 # shellcheck disable=SC2086
 docker-compose -f compose-backend.yml up -d $tostart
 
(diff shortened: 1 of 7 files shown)


jbygdell (approved):

.github/integration/setup/common/10_services.sh
```diff
@@ -14,19 +14,21 @@ if [ "$STORAGETYPE" = s3 ]; then
     tostart="mq db s3 certfixer"
 fi
 
+# we need to let it be split here otherwise it takes it as one
```
norling, line 17: This comment was a bit hard to read!
maybe something like
```suggestion
# We need to leave the $tostart variable unquoted here since we want it to split
```

norling (approved):
There's a comment that confused me for a second but in general this looks good to me! :)

teemukataja (approved):

a-ghanem:
I will try to do tests when we have the changes needed for TLS are incorporated in our CI/CD.

a-ghanem (approved):

blankdots:
merging as required number positive reviews achieved

PR #339, opened 2022-01-14
Refactor/handle improper uploads

aaperis:
Messages that concern uploads of wrongly encrypted files used to linger at the ingest queue. 

- These are now nacked and moved to the error queue, freeing up more consume resources and  improving overall performance of the ingest service.
- fixed some log formatting issues along the way

### DIFF
--- a/.github/integration/setup/common/22_create_and_encrypt_large_file.sh
+++ b/.github/integration/setup/common/22_create_and_encrypt_large_file.sh
@@ -17,3 +17,14 @@ touch empty.c4gh
 
 dd if=largefile.c4gh bs=1 count="$((15000+RANDOM))" of=truncated1.c4gh
 dd if=largefile.c4gh bs=1 count=10 of=truncated2.c4gh
+
+dd if=/dev/random of=wrongly_encrypted.raw count=1 bs=$(( 1024 * 1024 *  1 )) iflag=fullblock
+
+cat > wrong_key.pub <<EOF
+-----BEGIN CRYPT4GH PUBLIC KEY-----
+fGi6kvZlQN37PLmumygbNaSLf0NA+kj4KB1b70HFFCU=
+-----END CRYPT4GH PUBLIC KEY-----
+EOF
+
+crypt4gh encrypt --recipient_pk wrong_key.pub < wrongly_encrypted.raw > wrongly_encrypted.c4gh
+rm wrongly_encrypted.raw wrong_key.pub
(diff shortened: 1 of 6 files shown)


blankdots (commented):
can we have a test that simulates this, somewhere along the lines of wrong file uploaded check error queue that the message is there ?

a-ghanem (approved):

norling (approved):

a-ghanem (approved):
LGTM :)

blankdots (approved):
I approve but a test would be nice for this.

.github/integration/setup/common/22_create_and_encrypt_large_file.sh
```diff
@@ -17,3 +17,14 @@ touch empty.c4gh
...
 dd if=largefile.c4gh bs=1 count="$((15000+RANDOM))" of=truncated1.c4gh
 dd if=largefile.c4gh bs=1 count=10 of=truncated2.c4gh
+
+dd if=/dev/random of=wrongly_encrypted.raw count=1 bs=$(( 1024 * 1024 *  1 )) iflag=fullblock
+
+cat > wrong_key.pub <<EOF
+-----BEGIN CRYPT4GH PUBLIC KEY-----
+fGi6kvZlQN37PLmumygbNaSLf0NA+kj4KB1b70HFFCU=
+-----END CRYPT4GH PUBLIC KEY-----
+EOF
```
jbygdell, line 27: Why not use crypth4gh to generate a new keypair and use that public key?

aaperis: I wanted to avoid the hassle of using passphrases when generating a new keypair.

jbygdell: if you use the golang binary you can create the keypair like this:

```
crypt4gh generate -n c4gh -f wrong_key -p VoaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm
```

aaperis:
Now with tests. The `trigger_failures` script was turned off some time ago (see `exit 0` on top of file), so I left it that way.

.github/integration/setup/common/22_create_and_encrypt_large_file.sh
aaperis: nice! I was not using the golang binary indeed.  But it [seems](https://github.com/neicnordic/sda-pipeline/blob/ad9fa3d0da3b6aa86979c94c5e6ccb36af910f66/.github/integration/setup/common/20_tools.sh#L3) that the tests don't either. So maybe keep the current solution atm?

blankdots: you can generate keys using that python `crypt4gh` as well https://github.com/neicnordic/sda-helm/blob/master/dev_tools/scripts/create-secrets.sh#L25-L35 - I am fine with both generating a new one or having it hardcoded.

jbygdell (approved):

PR #348, opened 2022-02-04
Refactor/ingest nack checks

aaperis:
This PR aims to improve handling of messages that cause errors in ingest. 
Specifically,

- checks that requeued nacked messages will not clog the queue and adds a short comment 
for future reference  

- adds nacking mechanism to handle failure of opening a file (e.g. nonexistent files etc)

- refactors and fixes related trigger-failure integration tests (except from one). 
The tests are deactivated for now until #321 is closed.

- Closes #230 and partially fixes #321.

PR #353, opened 2022-02-15
Refactor/verify nack checks

### LINKED ISSUE #344: Check verify service Nacks for incorrect requeue
- [x] DB
- [x] MQ

aaperis:
This PR aims to improve handling of messages that cause errors in verify. 

- checked nacking mechanisms that are in place

- adds info to messages sent to the error queue so that their analysis is easier

- adds integration test for nacking of msgs that trigger "Getheader failure" (deactivated for now)

Closes #344

### DIFF
--- a/.github/integration/setup/common/22_create_and_encrypt_large_file.sh
+++ b/.github/integration/setup/common/22_create_and_encrypt_large_file.sh
@@ -34,3 +34,9 @@ EOD
 
 crypt4gh encrypt --recipient_pk wrong_key.pub < wrongly_encrypted.raw > wrongly_encrypted.c4gh
 rm -f wrongly_encrypted.raw wrong_key.pub wrong_key.key
+
+touch test_db_file.raw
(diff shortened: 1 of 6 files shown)


blankdots (approved):
looks sound.

Besides the similar comment as with the ingest nack, https://github.com/neicnordic/sda-pipeline/pull/348#pullrequestreview-874331158

I think we should also have a way to standardise the error messages to the broker struct, as `jsonError` struct  and `InfoError` struct seem to do the same thing minus the type of `OriginalMessage` and pretty sure there is a way to convert between those 2 types, unless we expect something else than JSON into that `OriginalMessage` - but this sounds like a different task than this.

aaperis:
I completely agree with both remarks. Opened #354 and #355 in this respect :-).

norling (approved):
looks good as far as I can tell! :)

dbampalikis (approved):

PR #368, opened 2022-03-10
Refactor error queue messages

### LINKED ISSUE #355: Standardize error queue messages
The error messages that pour in to the error queue should be made more uniform. This could be done e.g. by using one of the structs that we already have in place for errors.

aaperis:
This PR replaces all structs that were used to send messages to the error queue with one standard structure that is suitable for appending different kinds of info. In addition,
- sets output of json related errors to human readable
- fixes integration tests related to broken json errors (deactivated for now)

The logic behind using an empty interface as a holder for appended messages is to:
- allow for dynamic json so that when the file moves through the pipeline and relevant info accumulates, the exact message that triggered the error can be sent to the error queue
- handle different types, e.g. for broken or non json related errors where unmarshalling is not possible
- preserve the json format of the original message so that messages in the error queue are easy to parse, if needed

Closes #355

### DIFF
--- a/.github/integration/tests/common/10_trigger_failures.sh
+++ b/.github/integration/tests/common/10_trigger_failures.sh
@@ -24,10 +24,12 @@ function check_move_to_error_queue() {
 		RETRY_TIMES=$((RETRY_TIMES + 1))
 		if [ $RETRY_TIMES -eq 61 ]; then
 			echo "::error::Time out while waiting for msg to move to error queue, logs:"
-			echo
-			echo ingest
-			echo
-			docker logs --since="$now" ingest
+			for k in ingest verify; do
+				echo
+				echo "$k"
+				echo
+				docker logs --since="$now" "$k"
+			done
 			exit 1
 		fi
 		sleep 2
--- a/.github/integration/tests/common/10_trigger_failures.sh
+++ b/.github/integration/tests/common/10_trigger_failures.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
 
+if [ "$STORAGETYPE" = s3notls ]; then
+    exit 0
+fi
+
 cd dev_utils || exit 1
-exit 0
+
 #
 # Submit some messages that will trigger various failures. Do this
 # before the "real" work to verify that these failures are not top of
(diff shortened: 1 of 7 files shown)


blankdots (approved):

norling (approved):

.github/integration/tests/common/8_bad_messages.sh
```diff
@@ -6,10 +6,39 @@
 # queue and handled again and again.
 #
 
-# disable for now since this doesn't work properly
+# disable for now
```
jbygdell, line 9: If this test works, why is it still disabled (exit 0) on the next line

jbygdell (changes requested):
A test for `SendJSONError` is missing in the `broker_test.go` file, preferably with both json and string payload

aaperis:
Unit test for `SendJSONError` added in `broker_test.go`.

dbampalikis (approved):

jbygdell (commented):
one comment regarding the `10_trigger_failures` test

.github/integration/tests/common/10_trigger_failures.sh
```diff
@@ -24,10 +24,12 @@ function check_move_to_error_queue() {
...
-			echo
-			echo ingest
-			echo
-			docker logs --since="$now" ingest
+			for k in ingest verify; do
+				echo
+				echo "$k"
+				echo
+				docker logs --since="$now" "$k"
+			done
```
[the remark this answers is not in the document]
aaperis: yes, good point. `10_trigger_failures.sh` was left disabled until the posix part was fixed but there is no reason why it shouldn't be enabled for the storage types that it is working. Now it is enabled for s3.

.github/integration/tests/common/10_trigger_failures.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [ "$STORAGETYPE" = s3notls ]; then
+if [ "$STORAGETYPE" = s3notls ] || [ "$STORAGETYPE" = posix ]; then
```
jbygdell, line 3: Why this selection, there is nothing in this test that is specific to a deployment type.

aaperis: It could be that there is a different error message returned in one of the tests which causes the posix test to not pass.

kostas-kou (approved):

a-ghanem (approved):

Issue #268, opened 2021-06-16
Extend sda-sync service to also copy header


jbygdell:
For big picture we need to sync files between Sweden and Finland. To do this we need to be able to sync the header as well.

Extend the s3sync service so it can get the header from the database and reencrypt it using another public key.

Question about protocol? Should header and body be concatenated or should it be sent separately?

- The simplest solution would be to fetch the header, encrypt it and pass it through the MQ message

- [ ] reencrypt with Finish public key
- [ ] standalone deployment template

PR #391, opened 2022-04-14
Add header in the synced file

kostas-kou:
This PR extends the sync service from just copying the headless file from archive to get the header from the database, put it together with the file, decrypt and reencrypt with another public key.
This extension makes it easier for syncing files between Sweden and Finland.

Changes:
- An extra (new) key pair added
- New function for getting the header from database by using the stable id because sync service does not have the file id info
- Decrypts the header
- Reencrypts the header with the new public key
- Writes header to destination
- Added tests and nacks
- Changes in integration tests

Comments:
- The extension is implemented only for the S3 case for now
- The procedure for decrypting and reencrypting header has been done manually and the crypt4gh package is not used. This was done in another branch `feature/sync-header`. The problem there was that the decrypted binary test file `dummy_data.c4gh` had a very small difference in size when the file from the backup bucket was decrypted. I could not figure out what was wrong and that is why is done manually.

Closes[ #268](https://github.com/neicnordic/sda-pipeline/issues/268)


cmd/sync/sync.go
```diff
@@ -109,6 +123,23 @@ func main() {
 				message.AccessionID,
 				message.DecryptedChecksums)
 
+			// Get the header from db
+			header, err := db.GetHeaderStableId(message.AccessionID)
+			if err != nil {
+				log.Errorf("GetHeaderStableId failed "+
```
norling, line 129: There are a few error messages that are very similar to this. Might be worth looking into a helper function to format the repeating parts!

norling (approved):

kjellp:
Hi! i struggle to see how this improvement will influence the ingest pipeline from the current version of sda-pipeline? The described feature sounds very similar to what is currently happening as part of the sda-doa service (in java), that we use as part of the "export" pipeline. Will the ingest pipeline continue to work as before? I guess the sync service is not part of the ingest pipeline currently, and the changes will only affect LocalEGA setups using the sync service optionally?

viklund:
The `sync` service normally copies the unencrypted file body over to the backup store. We actually need that still in SE so this should be an option/setting. And yes it is quite similar to DOA but here it is for the case where the backup is to another "federated node" instead of to a local backup space. So it should be triggered by the same message that the current backup is triggered by.

Hopefully we can also use this as a base in the future for mirroring data between EGA nodes.

viklund (changes requested):
The header encryption should be optional and toggled by some setting.

cmd/sync/sync.go
kostas-kou: We discussed with Alex about error messages and nacks and we will add it probably for refactoring for all the services.

kostas-kou:
Sync header now is moved as an option and the default one is without the header (as it was before)

kjellp (commented):
The fixes to make the feature optional looks good to me. Don't have sufficient S3 knowledge to comment / approve on the more S3 specific parts.

dev_utils/config.yaml
```diff
@@ -23,6 +23,7 @@ backup:
   cacert: "./dev_utils/certs/ca.pem"
   # posix backend
   location: "dev_utils"
+  header: "false"
```
jbygdell, line 26:
```suggestion
  copyHeader: "false"
```
A more verbose name is preferable .

dev_utils/config.yaml
```diff
@@ -46,6 +47,7 @@ broker:
 c4gh:
   passphrase: "oaagCP1YgAZeEyl2eJAkHv9lkcWXWFgm"
   filepath: "./dev_utils/c4gh.sec.pem"
+  newpublickey: "./dev_utils/c4gh-new.pub.pem"
```
jbygdell, line 50: A better name for this should be though of, that is more descriptive other than just  `new`.

examples: 
* syncPubKey 
* backupPubKey

internal/config/config.go
```diff
@@ -359,3 +359,32 @@ func GetC4GHKey() (*[32]byte, error) {
...
+
+// CheckHeader reads the config and returns if the header will be copied
+func CheckHeader() bool {
+	header := viper.GetString("backup.header")
+	if header == "true" {
+		return true
+	} else {
+		return false
+	}
+}
```
jbygdell, line 390:
```suggestion
// CheckHeader reads the config and returns if the header will be copied
func CopyHeader() bool {
	if viper.IsSet("backup.copyHeader") {
		return viper.GetBool("backup.copyHeader")
	}
	
	return false
}
```

Viper can do direct string to boolean conversions for most cases true/false, 1/0 and so on.

internal/database/db.go
```diff
@@ -171,6 +171,21 @@ func (dbs *SQLdb) getHeader(fileID int) ([]byte, error) {
 	return header, nil
 }
 
+// GetHeaderStableId retrieves the file header by using stable id
+func (dbs *SQLdb) GetHeaderStableId(stableID string) (string, error) {
```
jbygdell, line 175:
```suggestion
func (dbs *SQLdb) GetHeaderForStableId(stableID string) (string, error) {
```
slightly more descriptive name

internal/database/db_test.go
```diff
@@ -306,6 +306,31 @@ func TestGetHeader(t *testing.T) {
 	log.SetOutput(os.Stdout)
 }
 
+func TestGetHeaderStableId(t *testing.T) {
```
jbygdell, line 309:
```suggestion
func TestGetHeaderForStableId(t *testing.T) {
```

internal/database/db_test.go
```diff
@@ -306,6 +306,31 @@ func TestGetHeader(t *testing.T) {
...
+	r := sqlTesterHelper(t, func(mock sqlmock.Sqlmock, testDb *SQLdb) error {
+
+		header := "0f40"
+		mock.ExpectQuery("SELECT header from local_ega.files WHERE stable_id = \\$1").
+			WithArgs("42").
+			WillReturnRows(sqlmock.NewRows([]string{"header"}).AddRow("0f40"))
+
+		x, err := testDb.GetHeaderStableId("42")
```
jbygdell, line 317:
```suggestion
		x, err := testDb.GetHeaderForStableId("42")
```

internal/config/config_test.go
```diff
@@ -646,3 +662,9 @@ func (suite *TestSuite) TestSyncConfiguration() {
 	assert.NoError(suite.T(), err)
 	assert.NotNil(suite.T(), config)
 }
+
+func (suite *TestSuite) TestCheckHeader() {
+	viper.Set("backup.header", "true")
+	cHeader := CheckHeader()
+	assert.Equal(suite.T(), cHeader, true, "The CheckHeader does not work")
+}
```
jbygdell, line 670:
```suggestion

func (suite *TestSuite) TestCopyHeader() {
	viper.Set("backup.CopyHeader", "true")
	cHeader := CopyHeader()
	assert.Equal(suite.T(), cHeader, true, "The CopyHeader check does not work")
}
```

cmd/sync/sync.go
```diff
@@ -295,6 +309,102 @@ func main() {
...
+							message.DecryptedChecksums,
+							e)
+					}
+				}
+
+				// write header to destination file
+				_, err = dest.Write(newHeader)
+				if err != nil {
+					log.Fatalf("Error in header writer: %v", err)
+				}
```
jbygdell, line 405: This will kill the app when triggered.
Instead a message should be sent like you do above

jbygdell (dismissed):

viklund (approved):
Since this is now an option, I approve. Jockes comments look good though.

dev_utils/config.yaml
kostas-kou: done

internal/config/config.go
kostas-kou: Cool, I did not know that and I did not want to put in config something else but strings. This solved it.

internal/config/config_test.go
kostas-kou: Done

cmd/sync/sync.go
kostas-kou: Right, I have missed it

jbygdell (approved):

blankdots (commented):
first a question does this work only with s3 ? i see no `posixheader` option, not sure it should work for :finland: :thinking: ?

kostas-kou:
As a first I focused on the s3 only but checking now for the posix

kostas-kou:
It is also ready for the posix.

kostas-kou:
It copies the header for the posix case as well

blankdots (approved):
The code looks ok, however I am missing how this would fit into the grand scheme of syncing. For example at the moment all it seems to do is re-encrypt the header, so i assume the file is still ingested through the same way (inbox > ingest etc.) which still needs the key of that archive, and i am not sure how re-encrypting would benefit in sync, or how is the actual file synced between nodes ?

Is the plan to spin 1 sync/back-up service that would archive the file to the back-up of the node as normal operation, and another 1 sync/back-up that would back-up to the other node ? 
can we have more details on the workflow ? - even thought the PR solves the issue it is missing context on how this would be operated.

I would prefer if we separate the back-up and the sync service into 2 separate service so it is easier to work with them in practice.

I approve, however please offer documentation as a follow-up how this would be used in practice.

PR #410, opened 2022-06-02
Fix integration test folder

dbampalikis:
This PR is a fix for the integration tests and specifically the `50_check_files.sh` file, and specifically when comparing checksums for the different types of backends. It includes:
- Add path to the certificates for all tls enabled backends
- Fail in case no accession ids are found in the integration test
- Change the db query for s3 backend

Closes: #411


kjellp (approved):
LGTM

a-ghanem (approved):

norling (approved):

PR #438, opened 2022-07-29
Notify service

### LINKED ISSUE #427: Create skeleton service for notify service
A simple go service skeleton for the service `notify`.

- Create it in the `cmd` subfolder
- It needs to recieve messages from the MQ, so a loop over the MQ is the base

jbygdell:
Some unit tests, not fully for smtp though
No integration tests

Closes #427


norling (approved):
Structure seems fine, but I'd like some description comments! :)

aaperis (approved):
Looks good  :-).  I agree with previous remark on comments. The `common` package could be quite useful for further shortening the code in the future. In its current form though, it would be nice if it included some short note that code is repeated.

internal/common/common.go
```diff
@@ -0,0 +1,48 @@
+package common
+
+import (
+	"github.com/xeipuuv/gojsonschema"
+)
+
+func ValidateJSON(reference string, body []byte) (*gojsonschema.Result, error) {
```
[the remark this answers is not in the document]
jbygdell: The function in the broker package will be moved into common later, then there will be no confusion anymore.
I don't wan't to put it in the PR as it is better to have that as a separate one.

cmd/notify/notify.go
```diff
@@ -0,0 +1,180 @@
...
+				continue
+			}
+			user := getUser(conf.Broker.Queue, d.Body)
+			if user == "" {
+				log.Errorln("No user in message, skipping")
+
+				continue
+			}
+
+			if err := sendEmail(conf.Notify, "THIS SHOULD TAKE A TEMPLATE", user, setSubject(conf.Broker.Queue)); err != nil {
```
blankdots, line 66: `"THIS SHOULD TAKE A TEMPLATE"` we leave like this or we configure an external file to read from?
should/can we use go templates for this ?

or we solve this in a different PR where we can add multiple templates depending on what happened per queue ?

jbygdell: The idea was to have a template file depending on  the reason for the email being sent, to be solved down the road in separate PRs.

blankdots: i am ok with this.

blankdots (approved):

viklund:
I think we can merge this now. It's mostly relevant for Fi and Se at the moment.

PR #441, opened 2022-08-15
Control API skeleton

### LINKED ISSUE #422: Create new go service skeleton for pipeline-api.
A simple go service skeleton for the service `pipeline-api`.

- Create it in the `cmd` subfolder
- It needs connection to the postgresql database
- It does NOT need to consume messages from the MQ
- It needs to send messages to the MQ
- The skeleton should include one `/health` API call that can be used by kubernetes (or docker swarm) to check that everything is ok.

jbygdell:
REST API skeleton

Needs the following to be usable:
* token validation (see `sda-download` for reference)
* DB queries to list uploaded files `(/files {"[ user | project ]": "NAME"})`
* DB queries to check ingestion status `(/ingestion/status {"project": "NAME"})`
* Start ingestion by sending messages `(/ingestion/start {"metadata": "METADATA_REFERENCE"})`

And maybe these as well: 
* configurable list of admin users that are allowed to query other users submission status

Closes #422


cmd/api/api.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	Conf.API.DB, err = database.NewDB(Conf.Database)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	go func() {
+		connError := Conf.API.MQ.ConnectionWatcher()
+		log.Error(connError)
+		os.Exit(1)
```
norling, line 41: It feels sort of scary to just exit here. Doesn't that mean that any mq connection error will kill the application regardless of the main thread state? Or can't this trigger once the connection is established?

norling (approved):
There are no function description comments, which is something I like! :)
Apart from that, it's fine by me!

cmd/api/api.go
jbygdell: It's what we currently use in all the other as well, but I agree we should probably replace this with a reconnect function.

.github/workflows/test.yml
```diff
@@ -35,8 +35,11 @@ jobs:
           bash ./make_certs.sh
           cd ..
 
+      - name: Start MQ and DB
+        run: docker-compose -f dev_utils/compose-no-tls.yml up -d db mq
+
       - name: Test
-        run: go test -v -coverprofile=coverage.txt -covermode=atomic ./...
+        run: go test --tags=integration -v -coverprofile=coverage.txt -covermode=atomic ./...
```
aaperis, line 42: is this tag needed?

aaperis (approved):
Looks good, as far as I can tell :-).

.github/workflows/test.yml
jbygdell: Yes, otherwise the tests against the live MQ server will not run.

aaperis: Yeah, figured it had something to do with that.  In my local ubuntu both tests run regardless of `--tags` but maybe this works differently in actions.

blankdots:
I am not 100% sure this closes #422 as there is a mention there for a `/health` endpoint and i probably missed that?

personally i see such an API endpoints secured either by a token or special headers, etc. - or will these be added by other PR's ?

cmd/api/api_integgration_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"sda-pipeline/internal/broker"
+	"sda-pipeline/internal/config"
+	"sda-pipeline/internal/database"
+
+	// "github.com/DATA-DOG/go-sqlmock"
```
blankdots, line 12: forgotten or to be used later ?

jbygdell:
> ~I am not 100% sure this closes #422 as there is a mention there for a `/health` endpoint and i probably missed that?~ - I see `/ready` is used instead - got me a bit confused
/ready is more applicable since it specifies if the service is ready to receive traffic

> 
> personally i see such an API endpoints secured either by a token or special headers, etc. - or will these be added by other PR's ?
Authentication would be required for the query endpoints, not for the status endpoint.
So the code for the authentication system is better together with PR that actually implements API endpoints that would require it.

blankdots (approved):
I am ok with this, however if we accept the move of sda-download here we will need to refactor it.

jbygdell:
> I am ok with this, however if we accept the move of `sda-download` here we will need to refactor it.

The refactoring will be both ways, since there will be quite a lot of code that will be shared between this service and `sda-download`

PR #447, opened 2022-09-06
Enforce dependency review not to update blindly

blankdots:
the question is if we refine to fail only on `critical` and `moderate` or we fail on all of them ?
see https://github.com/actions/dependency-review-action#configuration


jbygdell:
Definitively fail on `critical` and  `high`, maybe even on `moderate`

jbygdell (approved):

norling (approved):

PR #479, opened 2022-11-14
Docs/service descriptions

norling:
Adds markdown descriptions of all pipeline services (to be extended later with further information), as well as a trigger Action to update the documentation at neicnordic/neic-sda with the content of these files.

To make the action work, a secret needs to be added named `NEIC_SDA_ACCESS`, containing a github token with the
"Actions: Read and Write" permission on the neicnordic/neic-sda repository. Could someone with more permissions than me sort that out?

PR #485, opened 2022-11-23
Add case for backup in original location

dbampalikis:
The backup service is currently copying the file as in the same location as the archive (using the archive path). For the syncing between countries in BigPicture we need to copy the file in the original location (using the inbox path). 

This PR uses the original location in case the `copyHeader` parameter is set to true.

Co-Authored-By: kostas-kou <kkoumpouras@gmail.com>

### DIFF
--- a/.github/integration/tests/posixheader/50_check_files.sh
+++ b/.github/integration/tests/posixheader/50_check_files.sh
@@ -57,6 +57,11 @@ for aid in $accessids; do
 
 	rm -f "tmp/$apath"
 
+	apath=$(docker run --rm --name client --network dev_utils_default -v /home/runner/work/sda-pipeline/sda-pipeline/dev_utils/certs:/certs \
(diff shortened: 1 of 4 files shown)


jbygdell (approved):

norling (approved):

blankdots (commented):
seems ok. would be nice to document the expected workflow when the `copyheader` functionality is enabled. is this something that can be done as part of this PR or another PR ?

dbampalikis:
@blankdots sure I can do that! Should I add a README in the backup service folder or what do you have in mind?

blankdots:
> @blankdots sure I can do that! Should I add a README in the backup service folder or maybe in the main README file? Is that what you have in mind?

was thinking of something similar to what has been added in #479

dbampalikis:
yes that makes a lot of sense. I missed the #479 , maybe I should wait for it to be merged and update it there?

norling (approved):

dbampalikis:
> approved with the condition we add docs in another PR e.g. similar how it is done in #479

@blankdots I added some documentation, feel free to check it if you have the time :)

blankdots (approved):
lgtm

PR #486, opened 2022-11-25
[backup] make use of header re-encrypt

blankdots:
- crypt4gh package as of 1.6.0 has header and file re-encrypt functionality. - related PR: https://github.com/neicnordic/crypt4gh/pull/42

Not sure I see anything in the integration tests that the file and/or the header were re-encrypted ok, do we have something like this?


dbampalikis:
> 

Maybe I am off, but wouldn't [this](https://github.com/neicnordic/sda-pipeline/blob/6c23836fe8500d48237a99bb826dcf88ef03c0cb/.github/integration/tests/s3header/50_check_files.sh#L68) check that the file was actually correctly encrypted and decrypted with the new key?

norling (approved):
Code looks really neat! :)

dbampalikis (approved):
Looks good and reduces the code used for that by a lot!

blankdots:
> Maybe I am off, but wouldn't [this](https://github.com/neicnordic/sda-pipeline/blob/6c23836fe8500d48237a99bb826dcf88ef03c0cb/.github/integration/tests/s3header/50_check_files.sh#L68) check that the file was actually correctly encrypted and decrypted with the new key?

awesome, did not catch that. thx!

aaperis (approved):
looks nice :)

PR #487, opened 2022-11-25
Add nack when mapping fails

dbampalikis:
The PR fixes the bug with the mapper, which was acknowledging the message even in case of failure. The fix is focused on the sync functionality for the BigPicture, where the messages are all arriving in the same time to the destination, therefore the mapper might receive the message before the file has been ingested. Therefore, in case of failure, the mapper sleeps for up to 5 minutes, to give time to the database to recover or for the file to get ingested.

(Also, includes a couple of spelling mistakes in verify and db)

### DIFF
--- a/cmd/mapper/mapper.go
+++ b/cmd/mapper/mapper.go
@@ -5,7 +5,6 @@ package main
 import (
 	"encoding/json"
 	"os"
-	"time"
 
 	"sda-pipeline/internal/broker"
 	"sda-pipeline/internal/config"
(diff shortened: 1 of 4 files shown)


blankdots (approved):

norling (approved):

cmd/mapper/mapper.go
```diff
@@ -86,33 +86,47 @@ func main() {
...
 					"error: %v)",
-					d.CorrelationId,
+					delivered.CorrelationId,
 					mappings.DatasetID,
 					mappings.AccessionIDs,
 					err)
+
+				// Nack message so the server gets notified that something is wrong and requeue the message
+				if e := delivered.Nack(false, true); e != nil {
+					log.Errorf("Failed to nack following getheader error message "+
```
aaperis, line 96: The error message is not very clear. Maybe sth like:
```suggestion
					log.Errorf("Failed to nack message (failed to map files to dataset)"+
```
?

aaperis (approved):
The nack error message is unclear to me, but otherwise it looks good :)

PR #498, opened 2022-12-06
Feature/cancel messages

### LINKED ISSUE #484: Create a suggestion list on how to deal with "cancel messages"
Scenarios that can occur:
* cancel message arrives after ingest has started
* cancel message arrives before ingest has started
* cancel message arrives after ingest has completed
* ingest message arrives after file has been canceled

jbygdell:
POC of how to handle `cancel` messages from EGA.

`intercept` marks a file as `canceled` when a `cancel` message arrives, the message is then discarded.
On a `ingest` message `intercept` checks if the file exists in the DB, if not it inserts the file in the DB. If the file exists and has status `CANCELED` the status is set to `INIT`

`ingest` needs to check if a file has been marked as `canceled` before any work is started and before writing the state to the DB that signals that the work is completed.
* If ingestion has not started no further actions will be taken.
* If the file has gone through ingestion it will be removed from the archive.

`verify` needs to check if a file has been marked as `canceled` before any work is started and before writing the state to the DB that signals that the work is completed.
`verify` will no longer remove files from the inbox, this should be done only once a file has been given a stable ID.
* Since the file has gone through ingestion it will be removed from the archive.

- [x] cancel message arrives after ingest has started
- [x] cancel message arrives before ingest has started
- [x] cancel message arrives after ingest has completed
- [x] re-ingestion of a canceled file
- [x] Integration test for cancel messages

closes #484

### DIFF
--- a/.github/integration/tests/cancel/40_cancel_test.sh
+++ b/.github/integration/tests/cancel/40_cancel_test.sh
@@ -279,7 +279,7 @@ curl --cacert certs/ca.pem -u test:test 'https://localhost:15672/api/exchanges/t
 							}' | sed -e "s/DECMD5SUM/${decmd5sum}/" -e "s/DECSHA256SUM/${decsha256sum}/" -e "s/ACCESSIONID/${access}/")"
 
 RETRY_TIMES=0
-until docker logs finalize --since="$now" 2>&1 | grep "MarkReady failed"; do
+until docker logs finalize --since="$now" 2>&1 | grep "file is DISABLED, stopping work"; do
 	echo "case4 waiting for finalize to fail"
 	RETRY_TIMES=$((RETRY_TIMES + 1))
 	if [ "$RETRY_TIMES" -eq 60 ]; then
--- a/.github/integration/tests/cancel/40_cancel_test.sh
+++ b/.github/integration/tests/cancel/40_cancel_test.sh
@@ -145,3 +145,68 @@ until docker logs ingest --since="$now" 2>&1 | grep "file is DISABLED, reverting
 	fi
 	sleep 10
 done
+
+## case cancel message arrives before verification has completed
+
+docker pause verify
+
+now=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+curl --cacert certs/ca.pem -v -u test:test 'https://localhost:15672/api/exchanges/test/sda/publish' \
+	-H 'Content-Type: application/json;charset=UTF-8' \
+	--data-binary "$(echo '{"vhost":"test", "name":"sda",
+							"properties":{
+								"delivery_mode":2,
+								"correlation_id":"3001",
+								"content_encoding":"UTF-8",
+								"content_type":"application/json"
+							},
+							"routing_key":"files",
+							"payload_encoding":"string",
+							"payload":"{\"type\":\"ingest\",\"user\":\"test3\",\"filepath\":\"largefile.c4gh\",\"encrypted_checksums\":[
+								{\"type\":\"sha256\",\"value\":\"SHA256SUM\"},
+								{\"type\":\"md5\",\"value\":\"MD5SUM\"}
+							]}"
+							}' | sed -e "s/MD5SUM/${md5sum}/" -e "s/SHA256SUM/${sha256sum}/")"
+
+RETRY_TIMES=0
+until docker logs ingest --since="$now" 2>&1 | grep "Wrote archived file"; do
+	echo "waiting for ingestion to complete"
+	RETRY_TIMES=$((RETRY_TIMES + 1))
+	if [ "$RETRY_TIMES" -eq 60 ]; then
+		echo "::error::Time out while waiting for ingest to complete, logs:"
+		docker logs --since="$now" ingest
+		exit 1
+	fi
+	sleep 10
+done
+
+curl --cacert certs/ca.pem -v -u test:test 'https://localhost:15672/api/exchanges/test/sda/publish' \
+	-H 'Content-Type: application/json;charset=UTF-8' \
+	--data-binary "$(echo '{"vhost":"test", "name":"sda",
+							"properties":{
+								"delivery_mode":2,
+								"correlation_id":"3002",
+								"content_encoding":"UTF-8",
+								"content_type":"application/json"
+							},
+							"routing_key":"files",
+							"payload_encoding":"string",
+							"payload":"{\"type\":\"cancel\",\"user\":\"test3\",\"filepath\":\"largefile.c4gh\",\"encrypted_checksums\":[
+								{\"type\":\"sha256\",\"value\":\"SHA256SUM\"},
+								{\"type\":\"md5\",\"value\":\"MD5SUM\"}
+							]}"
+							}' | sed -e "s/MD5SUM/${md5sum}/" -e "s/SHA256SUM/${sha256sum}/")"
+
+docker unpause verify
+
+RETRY_TIMES=0
+until docker logs verify --since="$now" 2>&1 | grep "file is DISABLED, removing from archive"; do
+	echo "waiting for verify to abort"
+	RETRY_TIMES=$((RETRY_TIMES + 1))
+	if [ "$RETRY_TIMES" -eq 60 ]; then
+		echo "::error::Time out while waiting for verify to abort, logs:"
+		docker logs --since="$now" verify
+		exit 10
+	fi
+	sleep 10
+done
--- a/.github/integration/tests/cancel/40_cancel_test.sh
+++ b/.github/integration/tests/cancel/40_cancel_test.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+pip3 install s3cmd
+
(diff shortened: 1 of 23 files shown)


kjellp:
From FEGA Norway team discussions:
1) Important to think about the state of the file in the inbox as well (is it now deleted by default in the verify service after storing in the archive ?)  Relevant use case: end user deletes a run to correct the mapping between a Sample and a File, then the file must be in the inbox after the cancel, before the user adds the correct new Run/mapping.

2) A second ingest message/operation of a previously handled cancel must be possible: the canceled field of a file in the db, must somehow be reset before a second valid ingest of the same file. Could possibly be dealt with by the intercept service?

norling:
I think this breaks if an ingest message is followed by a cancel message, but the ingest message is then nack'ed and requeued. This will cause the cancel message to give an error when trying to update the database, but then the cancel message is still Ack'ed. Once the queue gets back to the requeued ingest message it's gonna be ingested right?

jbygdell:
> I think this breaks if an ingest message is followed by a cancel message, but the ingest message is then nack'ed and requeued. This will cause the cancel message to give an error when trying to update the database, but then the cancel message is still Ack'ed. Once the queue gets back to the requeued ingest message it's gonna be ingested right?

Ingest doesn't requeue on failure, it sends the message to the error queue.

norling:
yeah it does ^^ on line 264 of ingest, and 158 of intercept right? Also - I really think ingest should nack and requeue, at least on database errors where it makes sense that it could just resolve itself after waiting for some restart.

jbygdell:
> yeah it does ^^ on line 264 of ingest, and 158 of intercept right? Also - I really think ingest should nack and requeue, at least on database errors where it makes sense that it could just resolve itself after waiting for some restart.

True, it does requeue in cases where the error is local. But it will not ingest the file the next time it reads the message as the filestatus in the DB is `DISABLED`. The status only gets reset to `INIT` when `intercept` receives a new ingestion message from CEGA.

dbampalikis:
I actually think that what Martin is describing, is going to fail. If the `ingestion` message fails (for whatever reason) then the `cancel` message is also going to fail and get ack'ed. Then the `ingest` message that has been requeued is going to insert the file, which doesn't exist in the database, therefore leading to a file that hasn't been disabled.
Maybe I am missing something, but I think that the `cancel` message should only be ack'ed if i) the file exists and ii)is in specific status (probably `ready`, even if that means not unnecessary computation for a file that would be canceled), otherwise should be re-queued, until the file reaches this status.

jbygdell:
> I actually think that what Martin is describing, is going to fail. If the `ingestion` message fails (for whatever reason) then the `cancel` message is also going to fail and get ack'ed. Then the `ingest` message that has been requeued is going to insert the file, which doesn't exist in the database, therefore leading to a file that hasn't been disabled. Maybe I am missing something, but I think that the `cancel` message should only be ack'ed if i) the file exists and ii)is in specific status (probably `ready`, even if that means not unnecessary computation for a file that would be canceled), otherwise should be re-queued, until the file reaches this status.

Look at this in the full csequence.
* Ingest message comes from CEGA,

norling:
Now that you added the Nack I think it works! :) Gonna see if I can come up with any other weird cases ^^
In the mean time - please update the service descriptions with the changes you're making so that the documentation stays up to date!

dbampalikis:
One more case I figured out and we should take into account, is related to finalise. The scenario is:
- Ingestion message arrives and verify finishes and sends the message to CEGA
- Cancel message arrives and changes status to `DISABLED`
- Accession ids message arrives and tries to find a file with status `COMPLETED`, which doesn't exist, therefore finalise is requeing the message constantly.
I guess finalise should check for `DISABLED` status before trying to perform an action?

jbygdell:
> One more case I figured out and we should take into account, is related to finalise. The scenario is:
> 
> * Ingestion message arrives and verify finishes and sends the message to CEGA
> * Cancel message arrives and changes status to `DISABLED`
> * Accession ids message arrives and tries to find a file with status `COMPLETED`, which doesn't exist, therefore finalise is requeing the message constantly.
> 
> I guess finalise should check for `DISABLED` status before trying to perform an action?

Yes, that needs to be added.
This corner-case was unknown before the last FEGA-ops meeting where we got confirmation that the accessionID will always be sent even for files where the ingestion should be canceled.

dbampalikis:
Yeah I thought so, but I wanted to add it here so that we don't forget to add it

internal/database/db.go
```diff
@@ -465,3 +468,95 @@ func (dbs *SQLdb) Close() {
...
+		return err
+	}
+	if rowsAffected, _ := result.RowsAffected(); rowsAffected == 0 {
+		return errors.New("something went wrong with the query zero rows were changed")
+	}
+
+	return nil
+}
+
+// GetStatus returnes the status for a given fileID
```
norling, line 500:
```suggestion
// GetStatus returns the status for a given fileID
```

internal/database/db.go
```diff
@@ -465,3 +468,95 @@ func (dbs *SQLdb) Close() {
 	db := dbs.DB
 	db.Close()
 }
+
+// DisableFile marks the file as 'DISABLE'
```
norling, line 472:
```suggestion
// DisableFile marks the file as 'disabled'
```

internal/database/db.go
```diff
@@ -465,3 +468,95 @@ func (dbs *SQLdb) Close() {
...
+	const check = "SELECT status FROM local_ega.files WHERE id = $1;"
+	var status string
+	if err := db.QueryRow(check, fileID).Scan(&status); err != nil {
+		return "", err
+	}
+
+	return status, nil
+}
+
+// GetFileID returnes the ID for a given inbox path and user combination
```
norling, line 522:
```suggestion
// GetFileID returns the ID for a given inbox path and user combination
```

internal/database/db.go
```diff
@@ -465,3 +468,95 @@ func (dbs *SQLdb) Close() {
...
+		return 0, err
+	}
+
+	return fileID, nil
+}
+func (dbs *SQLdb) getFileID(filePath, user string) (int64, error) {
+	dbs.checkAndReconnectIfNeeded()
+
+	db := dbs.DB
+	const check = "SELECT id FROM local_ega.files WHERE inbox_path = $1 and elixir_id = $2;"
```
norling, line 535: There is no guarantee that this is a single unique file. In the case where the same user uploads a README for two projects this will match both of them... you probably need to add `archive_path` to make this unique

norling (changes requested):
damn it's hard to review things that are async :D 
I think there is a rare case of files being marked as disabled after the service checks status, but before the status is updated by the service which will result in a cancel message being functionally ignored. Do  you think that's worth dealing with? 
I think there's a case where the getFileID function isn't specific enough at least that needs to be fixed :)

internal/database/db.go
jbygdell: Well, that problem is not something we can do anything about since the message that comes from CEGA doesn't have an archive path for a file that is not ingested yet.

cmd/ingest/ingest.go
```diff
@@ -379,6 +401,33 @@ func main() {
...
+					log.Errorf("Failed to remove file from archive: %v", err)
+				}
+
+				if err := delivered.Nack(false, true); err != nil {
+					log.Errorf("Failed to nack message: %v", err)
+				}
+
+				continue
+			}
+			if status == "DISABLED" {
```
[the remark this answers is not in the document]
jbygdell: Highly unlikely, and the fix should be in the SetArchived call in that case.

.github/integration/tests/cancel/40_cancel_test.sh
```diff
@@ -0,0 +1,422 @@
...
+decsha256sum=$(crypt4gh decrypt --sk c4gh.sec.pem <largefile.c4gh | LANG=C dd bs=4M 2>"$dcf" | sha256sum | cut -d' ' -f 1)
+decmd5sum=$(crypt4gh decrypt --sk c4gh.sec.pem <largefile.c4gh | md5sum | cut -d' ' -f 1)
+
+rm -f "$dcf"
+
+## case1 cancel message arrives before ingestion has started
+
+docker pause ingest
+
+curl --cacert certs/ca.pem -u test:test 'https://localhost:15672/api/exchanges/test/sda/publish' \
```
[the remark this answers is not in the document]
jbygdell: I try to keep all scripts posix compatible, hence no functions.

cmd/verify/verify.md
```diff
@@ -45,4 +49,3 @@ Otherwise the processing continues with verification:
 
     1. The archive file is removed from the inbox storage.
```
blankdots, line 50: probably remove this information, if we are not doing it here anymore

blankdots (commented):
> verify will no longer remove files from the inbox, this should be done only once a file has been given a stable ID.

I am missing where this will be done. Or this is a separate PR ?
I assume `finalize` should be the one to do the file removal

PR #503, opened 2022-12-19
hotfix: [verify] send error message on decryption issues

jbygdell:
If verify can't decrypt the entire file we need to send an error message.


blankdots (approved):

norling (approved):
Great stuff! :) But please add this to the documentation as well, so that it stays up-to-date!

aaperis (approved):

Issue #412, opened 2022-06-03
Fix MQ connectivity


jbygdell:
If the connection to the MQ stalls, it will not be repoened unless the process is restarted.
relevant error mesage: `reason: Exception (504) Reason: \"channel/connection is not open\")"`

A/C: 
- [ ] create testcase for MQ reconnects
- [ ] check MQ connection before sending a message and reconnect if needed.

pontus:
Some comments in https://github.com/neicnordic/sda-pipeline/pull/517

Not sure what's going on here really (why the `NotifyClose`, e.g. https://github.com/neicnordic/sda-pipeline/blob/master/internal/broker/broker.go#L237 doesn't trigger). 

It doesn't seem likely, but possibly I could see this happen if the broker pod goes away (so the new one gets a new IP and nothing responds on the old IP) before the connection finally times out (after up to two hours).

jbygdell:
`NotifyClose` works for `connection` events but not for `channel` events, and those can not be tested either since there is no way to only close the channel from the MQ server side.

jbygdell:
The entire MQ handling really needs to be reworked from the bottom up in order to be able to solve the reconnect issue.

pontus:
For now, I mostly don't see why the `channel` goes wonky if the `connection` is still fine, so what's actually going wrong.

jbygdell:
A connection can have multiple channels

pontus:
Yes, so is this a RabbitMQ server bug then? :shrug: 

(To me it feels like any externally inflicted issue would be on the `connection` level.)

jbygdell:
The root problem is that as the code is written now we can't recreate a dropped channel from the services.

PR #517, opened 2023-01-24
Graceful shutdown

jbygdell:
* handle interrupt signals from the OS
* remove all `log.Fatal` to make sure we exit cleanly in all cases.
* linter fixes - return with no blank line before (nlreturn)

### DIFF
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -28,8 +28,7 @@ func main() {
 
 	Conf, err = config.NewConfig("api")
 	if err != nil {
-		log.Error(err)
-		sigc <- syscall.SIGINT
+		log.Fatal(err)
 	}
 	Conf.API.MQ, err = broker.NewMQ(Conf.Broker)
 	if err != nil {
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -93,9 +93,13 @@ func setup(config *config.Config) *http.Server {
 }
 
 func shutdown() {
-	defer Conf.API.MQ.Channel.Close()
-	defer Conf.API.MQ.Connection.Close()
-	defer Conf.API.DB.Close()
+	if Conf.API.MQ != nil {
+		Conf.API.MQ.Channel.Close()
+		Conf.API.MQ.Connection.Close()
+	}
+	if Conf.API.DB != nil {
+		Conf.API.DB.Close()
+	}
 }
 
 func readinessResponse(w http.ResponseWriter, r *http.Request) {
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -19,6 +19,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// Conf needs to be exported or the shutdown will not run properly
 var Conf *config.Config
 var err error
 
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -23,25 +23,29 @@ var Conf *config.Config
 var err error
 
 func main() {
+	sigc := make(chan os.Signal, 5)
+
 	Conf, err = config.NewConfig("api")
 	if err != nil {
-		log.Fatal(err)
+		log.Error(err)
+		sigc <- syscall.SIGINT
 	}
 	Conf.API.MQ, err = broker.NewMQ(Conf.Broker)
 	if err != nil {
-		log.Fatal(err)
+		log.Error(err)
+		sigc <- syscall.SIGINT
 	}
 	Conf.API.DB, err = database.NewDB(Conf.Database)
 	if err != nil {
-		log.Fatal(err)
+		log.Error(err)
+		sigc <- syscall.SIGINT
 	}
 
-	sigc := make(chan os.Signal, 5)
 	signal.Notify(sigc, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
 	go func() {
 		<-sigc
 		shutdown()
-		os.Exit(0)
+		os.Exit(1)
 	}()
 
 	srv := setup(Conf)
--- a/cmd/backup/backup.go
+++ b/cmd/backup/backup.go
@@ -7,7 +7,9 @@ import (
 	"encoding/json"
 	"io"
 	"os"
+	"os/signal"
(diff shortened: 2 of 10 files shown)


blankdots (approved):

pontus:
Looks good.

Some possible considerations:

* Would it make sense to log when exiting due to signal received (even though the "internal" signals might be confusing)?
* `cmd/api/api.go` has a few `log.Fatalln` - should those also be replaced?
* Maybe the signal handler is called enough that it's worth making it shared (supposedly being passwed mq, db or possibly `nil` if either isn't used)?
*

pontus (approved):

jbygdell:
> Looks good.
> 
> Some possible considerations:
> 
> * Would it make sense to log when exiting due to signal received (even though the "internal" signals might be confusing)?
> * `cmd/api/api.go` has a few `log.Fatalln` - should those also be replaced?

Since we call `shutdown()` befor calling `log.Fatalln` we ensure that the DB and MQ connections are closed cleanly
> * Maybe the signal handler is called enough that it's worth making it shared (supposedly being passwed mq, db or possibly `nil` if either isn't used)?

pontus (changes requested):
Now that I actually  manage to think; (generally) execution needs to be paused in the problematic thread as well.

Otherwise we risk e.g. failing to initialize configuration, noting that and requesting shutdown but still continue initialisation and getting a panic from using a nil.

jbygdell:
> Now that I actually manage to think; (generally) execution needs to be paused in the problematic thread as well.
> 
> Otherwise we risk e.g. failing to initialize configuration, noting that and requesting shutdown but still continue initialisation and getting a panic from using a nil.

Replacing the shutdown() and log.Fatalf() lines in `api.go` with a log.Errorf() and a SIGINT call will not change anything, since everything is already initiated at this point.

pontus:
> Replacing the shutdown() and log.Fatalf() lines in `api.go` with a log.Errorf() and a SIGINT call will not change anything, since everything is already initiated at this point.

Yes, that's more of a style/consistency issue whereas;

> > Now that I actually manage to think; (generally) execution needs to be paused in the problematic thread as well.
> > Otherwise we risk e.g. failing to initialize configuration, noting that and requesting shutdown but still continue initialisation and getting a panic from using a nil.
>

Is a parallelization issue - it seems certainly plausible that after signalling the shutdown (the channel has room), the thread were the error occurs goes on with the next thing and panics as previous initialization has failed before the goroutine responsible for shutdown gets a chance to read from the channel.

jbygdell:
Now it is nil pointer safe as well

pontus:
I'm not sure if it's unclear, but the problem I see (https://github.com/neicnordic/sda-pipeline/pull/517#pullrequestreview-1270540894) is a race condition, in short I think all instances where there are problems should be e.g.

```
		log.Error(err)
		sigc <- syscall.SIGINT
   	       <-make(chan bool, 1)
```

obviously, that requires moving at least the goroutine to do the actual shutdown up top.

It "probably" works anyway, but the way I see it that (or some other way of holding execution in the problematic flow) is needed for correctness.

(It may of course be that I'm wrong, but then it'd be nice to understand how.)

jbygdell:
> I'm not sure if it's unclear, but the problem I see ([#517 (review)](https://github.com/neicnordic/sda-pipeline/pull/517#pullrequestreview-1270540894)) is a race condition, in short I think all instances where there are problems should be e.g.
> 
> ```
> 		log.Error(err)
> 		sigc <- syscall.SIGINT
>    	       <-make(chan bool, 1)
> ```
> 
> obviously, that requires moving at least the goroutine to do the actual shutdown up top.
> 
> It "probably" works anyway, but the way I see it that (or some other way of holding execution in the problematic flow) is needed for correctness.
> 
> (It may of course be that I'm wrong, but then it'd be nice to understand how.)

The removal of `log.Fatal()` is to make sure that we have time to shutdown cleanly since it is the equivalent of doing log.Error() directly followed by os.Exit(1).

Utilizing the `syscall.SIGINT` is just a convenient way that also captures external shutdown requests, another way is to call a `shutdown()` function that would have to look something like this:
```
func shutdown(mq *broker.AMQPBroker, db *database.SQLdb) {
	if mq != nil {
		defer mq.Channel.Close()
		defer mq.Connection.Close()
	}
	if db != nil {
		defer db.Close()
	}
	os.Exit(1)
}

Given that all of this happens before the main work function is started it is pretty safe as is.

blankdots (approved):

aaperis:
The changes look reasonable to me but I would like to see some test cases, preferably in the form of code. This would help have a better picture of what to expect, given that we would like the pipeline's behavior to be as predictable as possible.

jbygdell:
> The changes look reasonable to me but I would like to see some test cases, preferably in the form of code. This would help have a better picture of what to expect, given that we would like the pipeline's behavior to be as predictable as possible.

Hard to have a coded test case that simulates an external `kill -2 PID`, especially since we can't test the main function as it is written now.
The closest we can get is to stop the container in the integration test and verify that we don't see any error in the DB or MQ logs `client unexpectedly closed TCP connection`

norling (approved):

blankdots (commented):
Does this fix https://github.com/neicnordic/sda-pipeline/issues/412 if not i suspect we might need to rewrite a bit some of the mq graceful shutdown for that as well.
there is an mention here: https://github.com/rabbitmq/amqp091-go/blob/main/doc.go#L107-L146

pontus:
I don't see that this should have any bearing on that - but then again, the suggested thing is already done and if an error is detected, the service bails out, so I'm not sure what happens when that stall occurs.

blankdots:
i did encounter it a couple of times, but was not able to reproduce it at all consistently.

https://github.com/rabbitmq/amqp091-go/pull/123 is the only related thing i have about it

blankdots:
the error looks like: https://github.com/streadway/amqp/issues/518

jbygdell:
Rebased on main

jbygdell:
Converting to draft since it probably will not have time to be implemented before we merge the repos.

jbygdell:
Will be implemented in the merged repo.

PR #526, opened 2023-02-08
mount uploads with named volume

### LINKED ISSUE #521: incorrectly mounted volumes
The volumes for the service `sftp-server` in the COMPOSE file `compose-sda.yml` is incorrectly mounted. 

The volume `uploads` should be mounted as named volume and `certs` should be used similar as in https://github.com/NBISweden/BigPicture-Deployment/blob/d4881abca46256de54837b0b50da1f25b26051f5/dev_env/docker-compose.yml#L261

nanjiangshu:
This PR closes #521, so that the stack can be run without the need to create a root folder `/uploads` on the system. 

After discussion, we decided not to change the `certfixer` since it will result in extensive modifications.


aaperis (approved):

blankdots (approved):
approved. 
conflict need to be addressed

dbampalikis (approved):
Agreed about the conflicts

PR #527, opened 2023-02-09
Update/sda mq

blankdots:
- small sanity update to make sure `sda-mq` still works in integration tests
- standard format the yaml files
- curl requests payload it is not interpreted properly with spaces and after so we strip the space.

### DIFF
--- a/dev_utils/compose-no-tls.yml
+++ b/dev_utils/compose-no-tls.yml
@@ -19,7 +19,7 @@ services:
     volumes:
       - /tmp/data:/data
   mq:
-    image: neicnordic/sda-mq:v1.4.24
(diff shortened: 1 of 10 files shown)


dev_utils/compose-no-tls.yml
```diff
@@ -19,18 +19,24 @@ services:
     volumes:
       - /tmp/data:/data
   mq:
-    image: neicnordic/sda-mq:v1.4.0
+    image: neicnordic/sda-mq:v1.4.24
```
jbygdell, line 22:
```suggestion
    image: ghcr.io/neicnordic/sda-mq:v1.4.24
```

just to keep it consistent

jbygdell (approved):

pontus (approved):

PR #535, opened 2023-02-22
switch to go 1.20 & prepare for repo merge #minor

blankdots:
This is in preparation for repository merging:

- this should be a version bump  v0.3.xxx to v0.4.0
- use consistent and stricter linting for golang files
- yaml config is pushed to 1.20 (sorry i use a yaml linter - in can revert the changes for that, but for me it makes it more consistent)
- update dependencies in one sweep and make package require golang 1.20
- there is a small bug with the tests that they don't run consistent, i suspect a timing issue so i added a sleep 20, but not 100% sure it will help long term.


jbygdell (approved):

aaperis (approved):

dbampalikis (approved):
LGTM. I am not merging yet in case someone else wants to check it

norling (approved):

PR #536, opened 2023-02-22
Update file if exists in ingest

dbampalikis:
The pull request is fixing an issue created by changes in the S3 Proxy. Specifically, the S3Proxy is now creating a row for each file that has been uploaded. That leads to ingest failing to insert the file, due to the unique constrains.
The PR checks the error when trying to insert the file and if it is related to the uniqueness constraint, it updates the file status instead.

### DIFF
--- a/internal/database/db.go
+++ b/internal/database/db.go
@@ -268,8 +268,19 @@ func (dbs *SQLdb) insertFile(filename, user string) (int64, error) {
 		"status, " +
 		"encryption_method) " +
 		"VALUES($1, $2, $3,'INIT', 'CRYPT4GH') RETURNING id;"
+
 	var fileID int64
+
 	err := db.QueryRow(query, filename, strings.ReplaceAll(filepath.Ext(filename), ".", ""), user).Scan(&fileID)
+	if err == nil {
+		return fileID, nil
+	}
+	// We can only handle the duplicate key error for now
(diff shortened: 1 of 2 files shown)


internal/database/db.go
```diff
@@ -266,11 +267,23 @@ func (dbs *SQLdb) insertFile(filename, user string) (int64, error) {
...
+			}
+
+			return fileID, nil
+		}
+
 		return 0, err
 	}
+
 	return fileID, nil
 }
```
viklund, line 288: I think it's difficult to read this many nested if statements. Why not flatten it like this:

```suggestion
    if err == nil {
        return fileID, nil
    }
    // We can only handle the duplicate key error for now
    if ! strings.Contains(err.Error(), "duplicate key value violates unique constraint") {
        return 0, err
    }

    const query = "UPDATE local_ega.main SET status = 'IN_INGESTION' WHERE local_ega.main.submission_file_path = $1 AND local_ega.main.submission_user = $2 AND local_ega.main.status = $3 RETURNING id;"
    err := db.QueryRow(query, filename, user, "INIT").Scan(&fileID)
    if err != nil {
        return 0, err
    }

    return fileID, nil
}
```

norling (approved):
I prefer @viklund's structure as well, looks good otherwise! :)

blankdots (approved):

aaperis (approved):
looks good! a small typo should be fixed before merging :)

PR #540, opened 2023-02-27
Feature/files api

dbampalikis:


internal/config/config.go
```diff
@@ -511,3 +520,36 @@ func CopyHeader() bool {
 
 	return false
 }
+
+// Function for reading the ega key in []byte
+func GetJwtKey(jwtpubkeypath string, jwtKeys map[string][]byte) error {
+	re := regexp.MustCompile(`(.*)\.+`)
```
jbygdell, line 526:
```suggestion
```

internal/config/config.go
```diff
@@ -511,3 +520,36 @@ func CopyHeader() bool {
...
+				if err != nil {
+					return fmt.Errorf("token file error: %v", err)
+				}
+				nameMatch := re.FindStringSubmatch(info.Name())
+
+				if nameMatch == nil || len(nameMatch) < 2 {
+					return fmt.Errorf("unexpected lack of substring match in filename %s", info.Name())
+				}
+
+				jwtKeys[nameMatch[1]] = keyData
```
jbygdell, line 545:
```suggestion
				jwtKeys[strings.TrimSuffix(info.Name(), filepath.Ext(info.Name()))] = keyData
```

jbygdell (commented):
Maybe not perfectly correct but you get the gists

cmd/api/api.go
```diff
@@ -124,7 +136,7 @@ func readinessResponse(w http.ResponseWriter, r *http.Request) {
 		statusCocde = http.StatusServiceUnavailable
 	}
 
-	w.WriteHeader(statusCocde)
+	c.JSON(statusCocde, "")
```
pontus, line 139: If we're rewriting this, could we untypo it (`statusCocde` -> `statusCode` unless I'm missing something)?

cmd/api/api.go
```diff
@@ -136,3 +148,126 @@ func checkDB(database *database.SQLdb, timeout time.Duration) error {
...
+	// Return response
+	c.JSON(200, files)
+}
+
+// getUserFromToken parses the token, validates it against the key and returns the key
+func getUserFromToken(w http.ResponseWriter, r *http.Request) (string, error) {
+	// Check that a token is provided
+	tokenStr, err := getToken(r.Header.Get("Authorization"))
+	if err != nil {
+		log.Error("authorisation header missing frm request")
```
pontus, line 183:
```suggestion
		log.Error("Authorization header missing from request")
```

Typo fix, use actual name of expected header.

cmd/api/api.go
```diff
@@ -5,16 +5,19 @@ import (
...
 	"syscall"
 	"time"
 
 	"sda-pipeline/internal/broker"
 	"sda-pipeline/internal/config"
 	"sda-pipeline/internal/database"
 
-	"github.com/gorilla/mux"
+	"github.com/gin-gonic/gin"
+	"github.com/golang-jwt/jwt"
```
blankdots, line 20: i recommend https://github.com/lestrrat-go/jwx it is better coverage and maintenance + we have it in sda-download already as we need jwk so it is better to have one library instead of multiple

internal/database/db.go
```diff
@@ -460,6 +466,62 @@ func (dbs *SQLdb) getArchived(user, filepath, checksum string) (string, int, err
...
+// getUserFiles is the actual function performing work for GetUserFiles
+func (dbs *SQLdb) getUserFiles(userID string) ([]*SubmissionFileInfo, error) {
+	dbs.checkAndReconnectIfNeeded()
+
+	files := []*SubmissionFileInfo{}
+	db := dbs.DB
+
+	const query = "SELECT inbox_path, status, created_at " +
+		"FROM local_ega.files " +
+		"WHERE elixir_id = $1;"
```
pontus, line 495:
```suggestion
	const query = "SELECT submission_file_path, status, created_at " +
		"FROM sda.files " +
		"WHERE submission_user = $1;"
```

cmd/api/api.go
dbampalikis: I am working on changing to that, not the nicest task, but I should be done soon :)

cmd/api/api.go
dbampalikis: Yeah that was a typo copy-pasted in multiple places. Thanks!

internal/database/db.go
dbampalikis: Right! Updated the query, although it's a bit more complicated than that, since the `sda.files` doesn't contain the status

cmd/api/api_test.go
```diff
@@ -1,27 +1,80 @@
...
+
+	WrongTokenAlgClaims = map[string]interface{}{
+		"iss":       "Online JWT Builder",
+		"iat":       time.Now().Unix(),
+		"exp":       time.Now().Add(time.Hour * 2).Unix(),
+		"aud":       "4e9416a7-3515-447a-b848-d4ac7a57f",
+		"sub":       "pleasefix@snurre-in-the-house.org",
+		"auth_time": "1632207224",
+		"jti":       "cc847f9c-7608-4b4f-9c6f-6e734813355f",
+	}
```
pontus, line 56: Unused?

pontus:
Looks OK, I notice one can have at most one key per issuer, meaning key rotation is not possible, but guess we won't have actual problems with that for now.

pontus (approved):

pontus:
I'm not sure exactly what use case this is meant to address, but I believe it would be nice to make it easier for `sda-cli` to support resuming uploads, in which case it would be nice to offer `submission_file_size` to the user as well if there is a checksum of type `UPLOADED`.

Or would that be out of scope?

cmd/api/api.go
```diff
@@ -94,11 +109,11 @@ func shutdown() {
...
 }
 
-func readinessResponse(w http.ResponseWriter, r *http.Request) {
-	statusCocde := http.StatusOK
+func readinessResponse(c *gin.Context) {
+	statusCode := http.StatusOK
 
 	if Conf.API.MQ.Connection.IsClosed() {
-		statusCocde = http.StatusServiceUnavailable
+		statusCode = http.StatusServiceUnavailable
```
blankdots, line 116: not sure i follow why we need MQ checks here,the API does not do anything with the broker.

cmd/api/api.go
```diff
@@ -136,3 +151,114 @@ func checkDB(database *database.SQLdb, timeout time.Duration) error {
...
+		c.JSON(500, err.Error())
+
+		return
+	}
+
+	// Return response
+	c.JSON(200, files)
+}
+
+// getUserFromToken parses the token, validates it against the key and returns the key
```
blankdots, line 181: i think the token check should be split, so that it is easier to integrate with sda download: see: https://github.com/neicnordic/sda-download/blob/main/pkg/auth/auth.go#L54-L90

Also it would be good to have a list of trusted issuers: https://github.com/neicnordic/sda-download/blob/main/pkg/auth/auth.go#L327-L344

internal/config/config.go
```diff
@@ -511,3 +519,28 @@ func CopyHeader() bool {
 
 	return false
 }
+
+// Function for reading the ega key in []byte
+func GetJwtKey(jwtpubkeypath string, jwtKeys map[string][]byte) error {
```
blankdots, line 524: i think we can have `/jwk` endpoint and serve our key there so it creates less hassle when changing the function for checking jwt signature.

blankdots (commented):
if i understand correctly the only way to match users to files is if we issue the token ... not sure how that will work from a services standpoint, for this reason I would prefer if having a local key for generating the token is not the default but an option as we can have a list of controlled issuers that we allow to generate the tokens and we can check the users against that.

internal/config/config.go
dbampalikis: Yes, that makes sense actually. I might need to leave this function as a second option though, at least until we create an endpoint for serving the key

pontus: It's not in https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml, but `.well-known/jwks.json` would be a good path for that, see e.g. https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets.

Issue #549, opened 2023-03-23
Add new messages from CEGA


jbygdell:
* Message `dataset release` - handled by mapper, marks files in a dataset as `ready` (https://github.com/EGA-archive/LocalEGA/blob/master/ingestion/schemas/dataset-release.json)
* Message `dataset deprecate` - handled by mapper, marks a dataset as `disabled` -- Needs testing ()https://github.com/EGA-archive/LocalEGA/blob/master/ingestion/schemas/dataset-deprecate.json
* Rename `accessionIDs` to `accession`

PR #552, opened 2023-03-29
Go implementation of the orchestrator, milestone PR to include functional code

kusalananda:
This PR adds the Go implementation of the orchestrator to the `sda-pipeline` repository.

The initial development of the code was provided in this branch by @blankdots two years ago, and the main changes I made to this code have been to bring it up to date with the current state of the `master` branch.

The purpose of this PR is to bring in the Go orchestrator code into the `master` branch and then to continue working on it to introduce manual checkpoints in the processing pipeline later. I _believe_ that the code is mostly a work-alike of the Python orchestrator found in the `sda-orchestrator` repository, although I have not done exhaustive testing of this.

The changes made to the _existing_ code in `sda-pipeline` includes changes to the `internal/config/config.go` code to support configuration variables specific to the orchestrator. At the moment, this is limited to a single variable, `PROJECT_FQDN`, which is used when creating accession and dataset IDs.

The creation of accession and dataset IDs is provisional in this PR and may well change completely later, but can be explained like this:

An ID is the URN representation of a SHA1 namespace UUID ("Verison 5 UUID") . The UUID is generated in a namespace given by another UUID generated from the hostname (fully qualified domain name) identifying the project, in combination with the body of the incoming message requesting the generation of the ID.

For an accession, this means that the generated ID is generated based on the project's FQDN and the `request` message, i.e., the username, file path, and the available hashes of the decrypted data (and the ordering of these pieces of information within the message).

The dataset ID is generated based on the project's FQDN and the `finalize` message, i.e., the same bits of information as for the accession ID, plus the accession ID itself.

In the future, I might add another orchestrator variable called `PROJECT_LABEL` that we can insert into the generated IDs to tell the IDs apart between projects visually. We may also want to distinguish accession IDs from dataset IDs(?)

NOTE: I'm a little bit confused why the dataset ID is computed _after_ the accession ID(s) in the pipeline, and why the `finalize` message type only allows for passing in a single accession ID while the response contains an array of accession IDs belonging to the dataset, but I'm assuming it has to do with reasons.  Or I've got it wrong along the way.

The code currently passes the integration test, which will be provided in another PR (#553).  That PR also contains the necessary environment file for the eth orchestrator, and the Docker compose file for standalone operation.  These things are not in this PR since they live under `dev_utils`.


cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,318 @@
...
+		case queueIngest:
+			publishMsg, publishType = ingestMessage(delivered.Body)
+		case queueMapping:
+			publishMsg, publishType = mappingMessage(delivered.Body, conf)
+		}
+
+		err = mq.ValidateJSON(&delivered,
+			routingSchema,
+			publishMsg,
+			publishType)
```
jbygdell, line 176: I really don't like this in my view unnecessary linesplitting since in many cases it makes the code harder to read.

```suggestion
		err = mq.ValidateJSON(&delivered, routingSchema, publishMsg, publishType)
```

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,318 @@
...
+
+		if err != nil {
+			log.Errorf("Validation of outgoing message failed "+
+				"(corr-id: %s, "+
+				"error: %v)",
+				delivered.CorrelationId,
+				err)
+
+			continue
+		}
```
jbygdell, line 186:
```suggestion
		if err != nil {
			log.Errorf("Validation of outgoing message failed, error: %v", err)
			if err := delivered.Nack(true); err != nil {
				log.Errorf("failed to nack message for reason: %v", err)
			}

			continue
		}
```

If this where to happen for some strange reason we need to `Nack` and re-queue the original message.

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,318 @@
...
+
+		log.Infof("Routing message "+
+			"(corr-id: %s, routingkey: %s, message: %s)",
+			delivered.CorrelationId,
+			routingKey,
+			publishMsg)
```
jbygdell, line 192:
```suggestion
		log.Debugf("Routing message (corr-id: %s, routingkey: %s, message: %s)",
			delivered.CorrelationId, routingKey, publishMsg)
```

jbygdell (changes requested):

cmd/orchestrate/orchestrate.go
kusalananda: I'm assuming the `Nack()` call should have been `Nack(false, true)`.

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+type trigger struct {
+	Type               string      `json:"type"`
+	User               string      `json:"user"`
+	Filepath           string      `json:"filepath"`
+	EncryptedChecksums []checksums `json:"encrypted_checksums"`
+}
+
+// finalize struct that holds the json message data
```
jbygdell, line 51:
```suggestion
```

Seems like a leftover that should be removed.

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,296 @@
...
+	"errors"
+	"fmt"
+	"os"
+
+	uuid "github.com/google/uuid"
+
+	"sda-pipeline/internal/broker"
+	"sda-pipeline/internal/config"
+
+	log "github.com/sirupsen/logrus"
```
jbygdell, line 16:
```suggestion

	"sda-pipeline/internal/broker"
	"sda-pipeline/internal/config"

	log "github.com/sirupsen/logrus"
	uuid "github.com/google/uuid"
```

gofmt should have caught this.

jbygdell (approved):

cmd/orchestrate/orchestrate.go
kusalananda: It didn't, but I'll fix it.

cmd/orchestrate/orchestrate.go
kusalananda: Sorted.

norling (approved):

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,314 @@
...
+// determined by the queue
+func schemaNameFromQueue(queue string, body []byte) (string, error) {
+	if queue == queueInbox {
+		return schemaFromInboxOperation(body)
+	}
+	m := map[string]string{
+		queueVerify:    "ingestion-accession-request",
+		queueComplete:  "ingestion-completion",
+		queueIngest:    "ingestion-trigger",
+		queueMapping:   "dataset-mapping",
```
blankdots, line 210: this is just a note: but with issue #549 we might need to change this

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,314 @@
...
+	return publish, new(finalize)
+}
+
+func mappingMessage(body []byte, conf *config.Config) ([]byte, interface{}) {
+	var message finalize
+	if err := json.Unmarshal(body, &message); err != nil {
+		return nil, nil
+	}
+	datasetID := uuid.NewSHA1(
+		uuid.NewSHA1(uuid.NameSpaceDNS, []byte(conf.Orchestrator.ProjectFQDN)),
```
blankdots, line 302: if we receive the Accession ids in a the `accession_ids` in a different order for the same message, will this generate the same id ?

maybe in a case where we an accession_id did not propagate soon enough and we need to replay the message will that give the same id

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,314 @@
...
+	"os"
+
+	"sda-pipeline/internal/broker"
+	"sda-pipeline/internal/config"
+
+	uuid "github.com/google/uuid"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
```
blankdots, line 18: would have preferred if these queues were configurable but the default was this, as this would enable for different deployment strategies

blankdots (changes requested):
~Would have preferred if the accession/dataset id generation are optional  in case an external service would like to generate them.~

Some small things to consider if the body has the keys in a different order (from what i understand there is no guarantee this would not happen).

Enable different queue configuration in case services are missing or some deployments would prefer to redirect messages to different queues

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,314 @@
...
+
+	return publish, new(finalize)
+}
+
+func mappingMessage(body []byte, conf *config.Config) ([]byte, interface{}) {
+	var message finalize
+	if err := json.Unmarshal(body, &message); err != nil {
+		return nil, nil
+	}
+	datasetID := uuid.NewSHA1(
```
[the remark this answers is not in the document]
blankdots: same as for accession id, no external services would be allowed

cmd/orchestrate/orchestrate.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func finalizeMessage(body []byte, conf *config.Config) ([]byte, interface{}) {
+	var message request
+	err := json.Unmarshal(body, &message)
+	if err != nil {
+		return nil, nil
+	}
+	accessionID := uuid.NewSHA1(
+		uuid.NewSHA1(uuid.NameSpaceDNS, []byte(conf.Orchestrator.ProjectFQDN)),
```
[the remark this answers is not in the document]
kusalananda: No, you would likely get a different UUID if the data is stored in a different order in the message.  Are you suggesting that it is necessary to be able to re-generate IDs in a reproducible way?

We are unlikely to use this _specific_ implementation of ID generation for accessions or datasets.  I believe the idea is to implement an ID service instead.  Each project that may want to use this service will likely want to tweak how IDs are generated and what they look like.  The point of this PR is instead to update your old code and get it to pass the integration tests before continuing to add manual "breakpoints" (points in the process where manual confirmation will be needed).

cmd/orchestrate/orchestrate.go
kusalananda: The response to the first part is the same as above. It is unclear whether reproducibility is an issue here. Of course, I could pick out specific parts of the message to generate the UUID, but in the end, this code will likely call an ID-generating service instead.

I do not understand the second part of the comment, sorry.

cmd/orchestrate/orchestrate.go
kusalananda: I can see what I can do about that in a future PR.

blankdots: ok, then i can wait for another PR about it

blankdots (commented):
i am ok with solving some of the issues mentioned in the comments in other PRs

blankdots (approved):

PR #554, opened 2023-03-31
error if file with same id exists

blankdots:
this is to prevent message requeuing so that the finalize gets blocked

- we check the id exists before we try to do anything with it and try to mark it as ready.
- add file for integration test (we skip for sftp header)
- update sda-db to v2.1.9

### DIFF
--- a/cmd/finalize/finalize.go
+++ b/cmd/finalize/finalize.go
@@ -146,6 +146,94 @@ func main() {
 				continue
 			}
 
+			accessionIDExists, err := db.CheckAccessionIDExists(message.AccessionID)
+			if err != nil {
+				log.Errorf("CheckAccessionIdExists failed "+
+					"(corr-id: %s, "+
+					"filepath: %s, "+
+					"user: %s, "+
+					"accessionid: %s, "+
+					"decryptedChecksums: %v, error: %v)",
+					delivered.CorrelationId,
+					message.Filepath,
+					message.User,
+					message.AccessionID,
+					message.DecryptedChecksums,
+					err)
+
+				if e := delivered.Nack(false, true); e != nil {
+					log.Errorf("Failed to NAck because of MarkReady failed "+
+						"(corr-id: %s, "+
+						"filepath: %s, "+
+						"user: %s, "+
+						"accessionid: %s, "+
+						"decryptedChecksums: %v, error: %v)",
+						delivered.CorrelationId,
+						message.Filepath,
+						message.User,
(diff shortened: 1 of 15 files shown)


blankdots:
related to https://github.com/neicnordic/sda-db/issues/83

pontus:
Looks to me like the logic is inverted (i.e. as you'd expect in C), am I just conused?

I know this is just a stop gap, would be glad if we got a test as well.

blankdots:
> Looks to me like the logic is inverted (i.e. as you'd expect in C), am I just conused?

i had an extra file, does this look better ?

will try to see a test

cmd/finalize/finalize.go
```diff
@@ -155,9 +156,13 @@ func main() {
 					message.AccessionID,
 					message.DecryptedChecksums,
 					err)
-
+				requeueMsg := true
+				if !strings.Contains(err.Error(), "duplicate key value violates unique constraint") {
+					requeueMsg = false
+				}
```
pontus, line 162: Ask not to requeue if the error isn't duplicate key - keep requeueing if it is?

blankdots: yes

blankdots: :facepalm: did not see the `!`

blankdots:
marked as draft till tests are done

internal/database/db.go
```diff
@@ -362,6 +362,45 @@ func (dbs *SQLdb) setArchived(file FileInfo, id int64) error {
...
+}
+
+// checkAccessionIdExists validates if an accessionID exists in the db
+func (dbs *SQLdb) checkAccessionIDExists(accessionID, filepath, checksum string) (bool, error) {
+	dbs.checkAndReconnectIfNeeded()
+
+	db := dbs.DB
+
+	const checkIDExist = "SELECT COUNT(stable_id) FROM local_ega.files WHERE stable_id = $1 " +
+		"and inbox_path = $2 and decrypted_file_checksum = $3 and status = 'READY';"
```
norling, line 390: Explicitly checking for path, checksum and staus might be a mistake here, as there could exist files with different value of these fields, but the same stable_id, and they'll still cause a unique key collision. Something like `SELECT COUNT(*) FROM sda.files WHERE stable_id = $1` is likely sufficient and safer. :)

norling (approved):

jbygdell (approved):

PR #556, opened 2023-04-04
Handle channel closures

### LINKED ISSUE #412: Fix MQ connectivity

jbygdell:
These seems to be the smallest changed needed in order to be able to handle a closed channel.

Using the library version of the Channel definition `*amqp.Channel` will require a much larger rewrite, specially of our tests.

Testing this is not straight forward as a channel can only be closed form the client side with the exemption of a channel time out event that happens after 30 min of inactivity.
To force a channel closure these actions needs to be taken:
* start MQ, DB and finalize
* pause the DB
* send a correct message to finalize
* wait for 30 min

closes #412

### DIFF
--- a/cmd/backup/backup.go
+++ b/cmd/backup/backup.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"io"
-	"os"
 	"strings"
 
 	"sda-pipeline/internal/broker"
@@ -35,6 +34,7 @@ type checksums struct {
 }
 
 func main() {
+	forever := make(chan bool)
 	conf, err := config.NewConfig("backup")
 	if err != nil {
 		log.Fatal(err)
@@ -78,10 +78,14 @@ func main() {
 	go func() {
 		connError := mq.ConnectionWatcher()
 		log.Error(connError)
-		os.Exit(1)
+		forever <- false
 	}()
 
-	forever := make(chan bool)
+	go func() {
+		connError := mq.ChannelWatcher()
+		log.Error(connError)
+		forever <- false
+	}()
 
 	log.Info("Starting backup service")
 	var message backup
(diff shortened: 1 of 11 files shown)


pontus:
Looks ok - a few general comments:

* Given that it's a failure if we exit, it would be nice to signal that, so a leading `defer func() { os.Exit(1) }()` would be nice
* As we're exiting the loop and have defers for closing, is there any benefit of closing the `db` on connection failure?
* I have a slight preference for exiting the loop directly on channel failure rather than closing the connection and waiting for that to be detected -  I guess either should work, though.

pontus (approved):

blankdots (approved):

blankdots (commented):
after this maybe we also need to rebase #517 ?

jbygdell:
> Looks ok - a few general comments:
> 
> * Given that it's a failure if we exit, it would be nice to signal that, so a leading `defer func() { os.Exit(1) }()` would be nice

It's not really an error, we just can't do a reconnect without restarting the main work loop.


> * As we're exiting the loop and have defers for closing, is there any benefit of closing the `db` on connection failure?
> * I have a slight preference for exiting the loop directly on channel failure rather than closing the connection and waiting for that to be detected -  I guess either should work, though.

jbygdell:
Let's see if this can be made somewhat cleaner, I found a way to reduce the timeout to be more manageable.

pontus:
I'd say seeing anything closed would imply an error (being why we normally loop forever), but I wouldn't have approved if I couldn't live with it, and don't really believe it matters.

blankdots (approved):

pontus (approved):

aaperis (approved):
looks good to me

PR #558, opened 2023-04-06
make orchestrator queues configurable

blankdots:
- it seems that orchestrator does not need a `broker.queue` by default
- making the queues configurable so that it allows for different deployment strategies (mentioned in: https://github.com/neicnordic/sda-pipeline/pull/552#pullrequestreview-1371233087)


kusalananda (approved):
This looks like a good improvement!

norling (approved):

PR #585, opened 2023-05-24
Allow for setting message prefetch number

### LINKED ISSUE #575: Ensure that services do not retrieve more messages than they can handle
Services can retrieve multiple messages at a time, making the time to ACK the later messages too long for rabbitmq. This needs to be limited to an amount that the service can be guaranteed to handle in time (during normal operation).
The amount of messages to retrieve should be configurable, with a default of 1.

**Additional notes**
Place to configure: in the client (GO code), in the connection part 

A/C

No integration test

jbygdell:
This fixes an issue when the queue contains to many messages and the time for processing each message takes so long that acking a message not being worked on times out after 30min.

- default number of messages to retrieve from the server is set to 2
- setting the prefetch number to 1 will allow for a true round-robin situation with multiple workers

Fixes #575

### DIFF
--- a/cmd/backup/backup.md
+++ b/cmd/backup/backup.md
@@ -48,6 +48,8 @@ These settings control how backup connects to the RabbitMQ message broker.
 
  - `BROKER_PASSWORD`: password to connect to rabbitmq
 
+ - `BROKER_PREFETCHCOUNT`: Number of messagfes to pull from the message server at the time (default to 2)
+
 ### PostgreSQL Database settings:
 
  - `DB_HOST`: hostname for the postgresql database
--- a/cmd/finalize/finalize.md
+++ b/cmd/finalize/finalize.md
@@ -36,6 +36,8 @@ These settings control how finalize connects to the RabbitMQ message broker.
 
  - `BROKER_PASSWORD`: password to connect to rabbitmq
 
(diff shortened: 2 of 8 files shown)


blankdots (approved):

pontus (approved):

dbampalikis (changes requested):
The implementation looks good. However, would it be possible to either add some documentation about it and possibly add it to the config (and or the docker compose file) so that we remember the option and what it is supposed to do?

jbygdell:
> The implementation looks good. However, would it be possible to either add some documentation about it and possibly add it to the config (and or the docker compose file) so that we remember the option and what it is supposed to do?

That will come in the helm chart update

dbampalikis:
> > The implementation looks good. However, would it be possible to either add some documentation about it and possibly add it to the config (and or the docker compose file) so that we remember the option and what it is supposed to do?
> 
> That will come in the helm chart update

It should be possible to set it on a docker level too, right? Why wouldn't we add it to the config, to know that this is an option then?

aaperis (approved):
It looks good :-). There's no harm in adding the configuration variable in the local dev environment though, as Dimi suggested. Regardless if that means repeating the default value, it may make life easier for future development.

dbampalikis (approved):

blankdots:
i merged this, looked like it was ready, sry if it shouldn't have been merged :(

PR #587, opened 2023-05-30
Handle cancel messages #minor

jbygdell:
Minimal set of changes needed to handle the cancel messages.
Affected services are:
- finalize
- ingest
- interceptor
- verify

Removal of files from the inbox has shifted from verify to mapper in order to be able to handle reingestion of a canceled file.


.github/integration/tests/cancel/40_cancel_test.sh
```diff
@@ -0,0 +1,480 @@
...
+
+C4GH_PASSPHRASE=$(grep -F passphrase config.yaml | sed -e 's/.* //' -e 's/"//g')
+export C4GH_PASSPHRASE
+
+ENC_MD5=$(md5sum largefile.c4gh | cut -d' ' -f 1)
+ENC_SHA=$(sha256sum largefile.c4gh | cut -d' ' -f 1)
+
+## check that all containers are running
+docker ps -a
+
```
pontus, line 17: Just for log output or should we react and bail out/wait if wrong?

blankdots:
we are using `disabled` for files under a dataset that has been deprecated, should we have maybe a different status than `disabled` not to get confused e.g. `cancelled` ?

or another question, can a file be `disabled` in a dataset and `enabled` in another one, would such a case occur ?

internal/database/db.go
```diff
@@ -194,41 +196,35 @@ func (dbs *SQLdb) GetHeaderForStableID(stableID string) (string, error) {
...
 	return err
 }
 
 // markCompleted performs actual work for MarkCompleted
-func (dbs *SQLdb) markCompleted(file FileInfo, fileID int) error {
+func (dbs *SQLdb) markCompleted(file FileInfo, fileID, corrID string) error {
 	dbs.checkAndReconnectIfNeeded()
 
 	db := dbs.DB
-	const completed = "UPDATE local_ega.files SET status = 'COMPLETED', " +
```
pontus, line 216: Maybe I should have thought of that in the DB rather PR, but is there any need for conversion here?

pontus: (of the database)

internal/database/db_test.go
```diff
@@ -192,104 +192,88 @@ func TestMarkCompleted(t *testing.T) {
...
 
-	buf.Reset()
+		i, err := testDb.GetFileID("f7207667-7d2d-46f5-96af-bd11416475c0")
+		assert.Equal(t, "0f83976fc-7e59-4a12-ad17-0154a36e36fc", i)
 
-	log.SetOutput(os.Stdout)
+		return err
+	})
+
+	assert.Nil(t, r, "GetMainID returned unexpected error")
```
pontus, line 265: `GetMainID`?

internal/database/db_test.go
```diff
@@ -592,6 +557,67 @@ func TestMapFilesToDataset(t *testing.T) {
...
+
+	err = sqlTesterHelper(t, func(mock sqlmock.Sqlmock, testDb *SQLdb) error {
+		mock.ExpectExec("INSERT INTO sda.file_event_log\\(file_id, event, correlation_id, user_id, message\\) VALUES\\(\\$1, \\$2, \\$3, \\$4, \\$5\\);").
+			WithArgs("bad-uuid", "error", "f83976fc-7e59-4a12-ad17-0154a36e36fc", "dummy", "").
+			WillReturnError(fmt.Errorf("error for testing"))
+
+		return testDb.UpdateFileStatus("bad-uuid", "error", "f83976fc-7e59-4a12-ad17-0154a36e36fc", "dummy", "")
+	})
+
+	assert.NotNil(t, err, "SetFileStatus did not fail as expected")
```
pontus, line 579:
```suggestion
	assert.NotNil(t, err, "UpdateFileStatus did not fail as expected")
```

pontus (approved):

.github/integration/tests/cancel/40_cancel_test.sh
jbygdell: Just a sanity check, the compose file has dependencies so that the container running the test scripts should wait for inbox to be ready before it starts.

internal/database/db.go
jbygdell: No, not since the query is switched from the legacy `local_ega` to the new `set_verified` function.

blankdots (approved):
small thing for consistency

we will need to update helm charts for verify and mapper
and my question above still stands, mostly because i don't understand when we will use `enable` in the db for a file, or i missed it. Also what would be the type of message that would process the `enabling` of a file/dataset

dbampalikis:
I haven't finished reviewing yet, but I have some first comments:
1. Could you update the documentation for the services, to include the cancel messages case?
2. Is the plan to get rid of the `sda-common` repo? I am asking since there are some functions that existed there (such as `GetVersion`) which are duplicated here
3. I am not sure about it, but should the minimum version be part of the config instead of hardcoded in the code?

aaperis (approved):
Looks great :-)

dbampalikis:
> > I haven't finished reviewing yet, but I have some first comments:
> > 
> > 1. Could you update the documentation for the services, to include the cancel messages case?
> 
> Wasn't that included? oh well, better look through the stashes then.
> 
> > 2. Is the plan to get rid of the `sda-common` repo? I am asking since there are some functions that existed there (such as `GetVersion`) which are duplicated here
> 
> `sda-common` has been archived and any duplications we be resolved when `sda-pipeline` is merged into the `sda` code in `sensitive-data-archive`

Ah I see, so we will fix all these ones the repos are merged, right?

> 
> > 3. I am not sure about it, but should the minimum version be part of the config instead of hardcoded in the code?
> 
> It's not configurable so it will have to be static regardless of where it is.
Maybe we should move it to the db, when it tries to connect initially?

dbampalikis (approved):
We went through the different cases with @aaperis and it looks good!

One corner case that we thought of (probably we don't need to solve this now, but just to have it in mind) is when the status of the files changes in verify, between getting the file status (line 126) and checking for disabled (line 147). So pretty much if verify finish the job while the query to the db is being executed. Again, it's a corner case, so maybe we don't need to care much about it?

pontus:
> One corner case that we thought of (probably we don't need to solve this now, but just to have it in mind) is when the status of the files changes in verify, between getting the file status (line 126) and checking for disabled (line 147). So pretty much if verify finish the job while the query to the db is being executed. Again, it's a corner case, so maybe we don't need to care much about it?

I think this is a valid consideration but fairly low risk, but if we want to address it, that should be done in `set_verified` over at https://github.com/neicnordic/sensitive-data-archive/blob/main/postgresql/initdb.d/02_functions.sql I think.

PR #588, opened 2023-05-30
Feature/new dataset messages #minor

blankdots:
This will tag a new #minor version
This requires: https://github.com/neicnordic/sda-db/pull/91 and new `sda` schema, so **it breaks backwards compatibility with old db versions**

Introduces:
- `finalize` only adds the accessionId to the file, it does not mark it as `ready` anymore 
- `mapper` 
   - `mapping` message assigns files to dataset
   - `release` message marks `files` as `ready` in the db
   - `deprecate` message marks the files as `disabled` in the db, added integration test for disabled
- `orchestrate` send also the release message (right now it waits for 1 minute)

#### Mention

- other changes will come to `sda-orchestrator` (python version) and `sda-download` in different PRs
- ~last commit should be removed once db PR it is approved~

### DIFF
--- a/cmd/intercept/intercept.go
+++ b/cmd/intercept/intercept.go
@@ -18,6 +18,8 @@ const (
 	msgCancel    string = "cancel"
 	msgIngest    string = "ingest"
 	msgMapping   string = "mapping"
+	msgRelease   string = "release"
+	msgDeprecate string = "deprecate"
 )
 
 func main() {
@@ -125,6 +127,8 @@ func main() {
 				msgAccession: "accessionIDs",
 				msgIngest:    "ingest",
 				msgMapping:   "mappings",
+				msgRelease:   "mappings",
+				msgDeprecate: "mappings",
 			}
 
 			routingKey := routing[msgType]
@@ -159,6 +163,8 @@ func schemaNameFromType(msgType string) (string, error) {
 		msgCancel:    "ingestion-trigger",
 		msgIngest:    "ingestion-trigger",
 		msgMapping:   "dataset-mapping",
+		msgRelease:   "dataset-release",
+		msgDeprecate: "dataset-deprecate",
 	}
 
 	if m[msgType] != "" {
--- a/cmd/mapper/mapper.go
+++ b/cmd/mapper/mapper.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 
 	"sda-pipeline/internal/broker"
 	"sda-pipeline/internal/config"
@@ -59,7 +60,23 @@ func main() {
 		}
 		for delivered := range messages {
 			log.Debugf("received a message: %s", delivered.Body)
-			err := mq.ValidateJSON(&delivered, "dataset-mapping", delivered.Body, &mappings)
+
+			schema, err := schemaFromDatasetOperation(delivered.Body)
+
+			if err != nil {
+				log.Errorf(err.Error())
+
+				if err := delivered.Ack(false); err != nil {
+					log.Errorf("failed to ack message: %v", err)
+				}
+				if err := mq.SendMessage(delivered.CorrelationId, mq.Conf.Exchange, conf.Broker.RoutingError, conf.Broker.Durable, delivered.Body); err != nil {
+					log.Errorf("failed to send error message: %v", err)
+				}
+
+				continue
+			}
+
+			err = mq.ValidateJSON(&delivered, schema, delivered.Body, &mappings)
 			if err != nil {
 				log.Errorf("Failed to validate message for work "+
 					"(corr-id: %s, "+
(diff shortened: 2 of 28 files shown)


cmd/intercept/intercept.go
```diff
@@ -125,6 +127,8 @@ func main() {
 				msgAccession: "accessionIDs",
 				msgIngest:    "ingest",
 				msgMapping:   "mappings",
+				msgRelease:   "mappings",
+				msgDeprecate: "mappings",
```
teemukataja, line 131: `msgRelease` and `msgDeprecate` should probably be something different than `mappings`, but not sure what, since there are similar things many times (singular, plural, prefix-name).

If I understand correctly, we should use these constants https://github.com/neicnordic/sda-pipeline/blob/d4eefc99c44652efe1ea61bde9ae95a5012de496/cmd/intercept/intercept.go#L16-L23 instead of building the dictionary here https://github.com/neicnordic/sda-pipeline/blob/d4eefc99c44652efe1ea61bde9ae95a5012de496/cmd/intercept/intercept.go#L126-L132 and here https://github.com/neicnordic/sda-pipeline/blob/d4eefc99c44652efe1ea61bde9ae95a5012de496/cmd/intercept/intercept.go#L161-L168

teemukataja, line 131:
```suggestion
				msgRelease:   "release",
				msgDeprecate: "deprecate",
```
`msgRelease` and `msgDeprecate` should probably be `release` and `deprecate`, but not sure, since there are similar things many times (singular, plural, prefix-name).

If I understand correctly, we should use these constants https://github.com/neicnordic/sda-pipeline/blob/d4eefc99c44652efe1ea61bde9ae95a5012de496/cmd/intercept/intercept.go#L16-L23 instead of building the dictionary here https://github.com/neicnordic/sda-pipeline/blob/d4eefc99c44652efe1ea61bde9ae95a5012de496/cmd/intercept/intercept.go#L126-L132 and here https://github.com/neicnordic/sda-pipeline/blob/d4eefc99c44652efe1ea61bde9ae95a5012de496/cmd/intercept/intercept.go#L161-L168

cmd/mapper/mapper.go
```diff
@@ -137,3 +222,35 @@ func main() {
...
+		return "", errors.New("Malformed message, dataset message type is missing")
+	}
+
+	datasetOpsType, ok := datasetMessageType.(string)
+	if !ok {
+		return "", errors.New("Could not cast operation attribute to string")
+	}
+
+	switch datasetOpsType {
+	case "mapping":
```
teemukataja, line 246: `mapping` or `mappings`

cmd/mapper/mapper.go
```diff
@@ -84,41 +101,109 @@ func main() {
...
-					"(corr-id: %s, "+
-					"datasetid: %s, "+
-					"accessionids: %v, "+
-					"error: %v)",
-					delivered.CorrelationId,
-					mappings.DatasetID,
-					mappings.AccessionIDs,
-					err)
+			switch mappings.Type {
+			case "mapping":
```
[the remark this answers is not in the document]
blankdots: https://github.com/EGA-archive/LocalEGA/blob/master/ingestion/schemas/dataset-mapping.json#L18 it is `mapping` the message type, `mappings` is the queue name: https://github.com/neicnordic/sda-pipeline/blob/master/dev_utils/env.mapper#L3

cmd/intercept/intercept.go
blankdots: nope this tells to which queue the type of message: `mapping`, `release` and `deprecate` should go to, which is the `mappings` queue

cmd/mapper/mapper.go
blankdots: `mapping` as it is not queue name `mappings`

blankdots: this gets the json schema corresponding with the type found in the message

cmd/intercept/intercept.go
blankdots: no, those consts are used to map message types with json schema corresponding in the dictionary, I am not sure exactly why it was built like that, and that is not the scope of this issue, just following what was done.

teemukataja (approved):

jbygdell (approved):

cmd/orchestrate/orchestrate.go
```diff
@@ -157,33 +159,51 @@ func processQueue(mq *broker.AMQPBroker, queue string, routingKey string, conf *
...
+			if err != nil {
+				log.Errorf("Validation of outgoing message failed, error: %v", err)
+				if err := delivered.Nack(false, true); err != nil {
+					log.Errorf("failed to nack message for reason: %v", err)
+				}
+
+				continue
+			}
+			// let us wait a minute before sending the release message
+			time.Sleep(1 * time.Minute)
```
norling, line 194: Why do we wait a minute? Should it be configurable?

internal/database/db_test.go
```diff
@@ -456,65 +513,67 @@ func TestMapFilesToDataset(t *testing.T) {
...
 			err := testDb.MapFilesToDataset(di, acs)
 
 			if !assert.Nil(t, err, "MapFilesToDataset failed unexpectedly") {
 				return err
 			}
 		}
 
-		var buf bytes.Buffer
-		log.SetOutput(&buf)
+		// var buf bytes.Buffer
```
norling, line 547: I'm not a huge fan of keeping commented out code. Would it be ok to remove it instead?

norling (approved):
minor comments, but looks good in general! :)

cmd/orchestrate/orchestrate.go
blankdots: just as default, i can make it configurable

internal/database/db_test.go
blankdots: i need to redo that test ... i missed it

blankdots:
need to redo a test putting it to draft ...

cmd/orchestrate/orchestrate.go
```diff
@@ -302,3 +356,42 @@ func mappingMessage(body []byte, conf *config.Config) ([]byte, interface{}) {
...
+
+	publish, _ := json.Marshal(&msg)
+
+	return publish, new(mapping)
+}
+
+func validateMsg(delivered *amqp091.Delivery, mq *broker.AMQPBroker, routingKey string, durable bool, routingSchema string, publishMsg []byte, publishType interface{}) error {
+	err := mq.ValidateJSON(delivered, routingSchema, publishMsg, publishType)
+	if err != nil {
+		return nil
```
pontus, line 382: Is this correct ("invert error state")? If so, could we have a comment about why?

blankdots: that looks like a miss on my part, good catch fixing ...

